PR #13, opened 2019-02-18
Add style prop to Flex and FlexItem

lzhang-cbi:
Add style prop to Flex and FlexItem to allow dynamic styles to the components like height.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ This is an example of a brief overview of the _Major_ or _Minor_ version changes
 
 ---
 
+## [2.21] flex style
+
+#### **Changes**
+- Adds `style` prop to `Flex` and `FlexItem`.
+
 ## [2.2] zindex vars
 
 #### **Changes**
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,7 +7,7 @@
 * Jenkins populates the patch version depending on the branch.
 */
 
-String VERSION = "2.2"
+String VERSION = "2.21"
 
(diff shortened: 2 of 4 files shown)


akdetrick:
Thanks for your contribution!

For this change, I recommend against allowing extra inline styles to be passed in. To add custom styles, developers should use a separate element inside the `Flex` or `FlexItem`.

For example:
```jsx
<Flex>
	<FlexItem shrink>
		<div style={ /* custom styles go here */ }> ...some content... </div>
	</FlexItem>
</Flex>
```

For low-level components like these, it's best that they maintain a single responsibility. In this case, the components are responsible for one and only one thing: creating flexbox layouts that work correctly in every supported browser. If extra styles are passed in, they could conflict with the base styles in this component.

Basically, we never want to end up in a situation where the documentation is telling a lie about what a given `Flex` or `FlexItem` is actually doing in the app. They should behave the same way everywhere.

akdetrick:
If the above example of using a separate element for custom styles will fit your use case, we can close this PR. Let me know

lzhang-cbi:
I thought having the style prop would be the same as having className prop but allowing more dynamic styles.  It wouldnt affect other Flex and FlexItem components that dont have the style passed in.  
The above example works, I just dont like adding extra divs because im looping through them and itll create too many divs.  We can close the PR.

akdetrick:
Good point about the `className` prop. I probably shouldn't have included that to discourage overriding styles for Flex and FlexItem.

It might be a good idea to remove that `className` prop, but it would be a breaking change - I'm not sure if it's worth it yet

PR #50, opened 2019-07-08
added ButtonGroup and GroupButton

rey-wright:
## Packages changed

- [ ] `fds-dictionary`
- [ ] `fds-styles`
- [X] `fds-components`
- [ ] `fds-mui-theme`
- [ ] `fds-icons`

Added `GroupButton` and `ButtonGroup`

## Screenshots
Screen Shot 2019-07-08 at 3 07 43 PM

## Checklist
- [X] All unit tests pass


packages/fds-components/src/components/buttons/GroupButton.jsx
```diff
@@ -0,0 +1,32 @@
...
+    <button className={rootClass} {...rest}>
+      {Icon && <Icon size="xs" className="margin--right--half" />}
+      {children}
+    </button>
+  );
+};
+
+GroupButton.propTypes = {
+  /** Displays active button style when true */
+  active: PropTypes.bool,
```
akdetrick: For boolean props in FDS, I've been using `is`/`has` conventions as an extra hint. In this case, it would be `isActive`

packages/fds-components/src/components/buttons/GroupButton.md
```diff
@@ -0,0 +1 @@
+`<GroupButton />` should only be used inside `<ButtonGroup />`. This component controls it's own state. See above for example usage.
```
akdetrick: `GroupButton` is stateless, so part of this statement may be misleading

rey-wright: You think I should just leave off the 2nd sentence? Essentially what I'm trying to communicate, is that ButtonGroup doesn't control any active states for the buttons inside of it... they control their own. That being said, maybe this is obvious looking at the example.

packages/fds-components/src/components/buttons/GroupButton.jsx
```diff
@@ -0,0 +1,32 @@
...
+  const { className, active, Icon, children, ...rest } = props;
+
+  const rootClass = cx('group-btn', className, {
+    active,
+  });
+
+  return (
+    <button className={rootClass} {...rest}>
+      {Icon && <Icon size="xs" className="margin--right--half" />}
+      {children}
```
akdetrick: Using `children` for the label provides more formatting flexibility for the label.
Using a `label` prop would give the component more control/restriction over how the label is displayed.

One nice thing about restricting this down to a `label: PropTypes.string` prop is that we don't have to account for accidental nesting.

rey-wright: Two questions:

1. If there was a label prop., would we allow children?
2. What would you think about blending the two strategies, and requiring children to be a string? There's arguments against that, but wondering what you think.

packages/fds-components/src/components/buttons/GroupButton.md
rey-wright: I've gone ahead and removed the 2nd sentence.

packages/fds-components/src/style/buttons/buttonGroup.css
```diff
@@ -0,0 +1,70 @@
+:root {
+  --button-group-timeout: 250ms;
+}
+
+.btn-group {
+  display: inline-flex;
+}
+
+.btn-group > .group-btn {
```
akdetrick: We probably don't need these direct child combinators for buttons that are ancestors of `btn-group`.

If someone is deeply nesting buttons inside buttons, they're doing something wrong. We may also be able to guarantee ancestry if we want to go with a `label` prop on `GroupButton` instead of using `children`.

packages/fds-components/src/style/buttons/buttonGroup.css
```diff
@@ -0,0 +1,70 @@
...
+  margin-left: -1px;
+  position: relative;
+  border-radius: 0;
+}
+
+.btn-group > .group-btn:focus {
+  z-index: 2;
```
akdetrick: 🆒

packages/fds-components/src/components/buttons/GroupButton.jsx
rey-wright: I've gone ahead and opted for label, following our "lock it down by default" sort of strategy.

packages/fds-components/src/style/buttons/buttonGroup.css
```diff
@@ -0,0 +1,70 @@
+:root {
+  --button-group-timeout: 250ms;
+}
+
+.btn-group {
```
akdetrick: *Not a blocker*

I've been using hyphens as a convention to denote a `parent-child` relationship. We can always go back and change these later after we discuss what kind of naming conventions we want to use going forward

packages/fds-components/src/components/buttons/GroupButton.jsx
rey-wright: But yeah I think it's a strong idea 👍

packages/fds-components/src/style/index.css
```diff
@@ -5,6 +5,7 @@
 @import "fds-icons/lib/fds-icons";
 
 /* Components */
+@import "./buttons/buttonGroup.css";
```
akdetrick: Do you think this should go in `interactive/buttonGroup`, or buttons should be their own category?

If we want to make a new category for buttons, I can file a ticket to reclassify the stuff in "interactive".

packages/fds-components/src/style/buttons/buttonGroup.css
```diff
@@ -0,0 +1,70 @@
...
+.group-btn:focus {
+  background-color: #D3DDE4;
+}
+
+.group-btn:hover {
+  text-decoration: none;
+  background-color: #DEE5EB;
+}
+
+.group-btn.active {
```
akdetrick: Should active buttons have bold text?

packages/fds-components/src/components/buttons/GroupButton.jsx
akdetrick: I feel like it's more of an all or nothing decision - accept any children, or accept only one primitive to lock it down (string, number, etc).

One more argument _for_ using a dedicated prop instead of children is that props are self-documenting with prop types. It's nice to give hints.

packages/fds-components/src/style/buttons/buttonGroup.css
rey-wright: According to Sketch they should. Good catch.

Though one thing I worry about here: if we bold the text, the buttons shift ever so slightly, because bold makes the font width increase. Are we ok with that?

packages/fds-components/src/style/buttons/buttonGroup.css
rey-wright: If you have a pre-existing convention, I'd rather follow it 🙂 are you ok with .btngroup and .groupbtn?

packages/fds-components/src/style/index.css
rey-wright: I actually didn't notice the interactive. I think that makes sense to go there 👍

rey-wright:
I forgot to add changelog... adding now.

rey-wright:
Hey @akdetrick ... so I got to thinking after you left... this idea of keeping the interface rigid if we don't need the flexibility.

So here's the current way you use this component:

Screen Shot 2019-07-09 at 7 01 53 PM

I then decided to create a much more simpler API:

Screen Shot 2019-07-09 at 7 01 17 PM

This works exactly the same as the first version, is much less code / wiring up, and eschews the extra ButtonGroup component (I fully coded it up). I'm almost leaning towards this version but wanted to get your thoughts (and also asked Steven for his thoughts).

rey-wright:
Screen Shot 2019-07-09 at 7 27 54 PM

This could also be a really nice middle ground between those two states that I posted.

rey-wright:
I've gone ahead and changed the api to the last version I posted. Seems to be more rigid, but still flexible.

packages/fds-components/src/components/buttons/ButtonGroup.md
```diff
@@ -0,0 +1,27 @@
...
+  { value: 'Table', icon: TableIcon},
+  { value: 'Map', icon: MarketMapIcon},
+  { value: 'Suggested', icon: StarEmptyIcon, onClick: custom },
+];
+
+const [firstButton] = buttons;
+
+const [activeValue, setActiveValue] = useState(firstButton.value);
+
+<ButtonGroup buttons={buttons} activeValue={activeValue} onChange={setActiveValue} />
```
akdetrick: When the requirements are simple, I really like this approach of passing a simple config into a prop instead of requiring extra components. While thinking this over, I checked the current implementation and it looks like we also must support a "disabled" state.

`activeValue` sets a single active value, which makes this work, but how would we handle disabled states given more than one button could be disabled?

akdetrick: Sorry I didn't catch that requirement earlier. There may be an argument for the `GroupButton` component again as a declarative way of setting active/disabled as opposed to the config object and manager pattern

packages/fds-components/src/components/buttons/ButtonGroup.jsx
```diff
@@ -0,0 +1,47 @@
...
+    </div>
+  );
+};
+
+ButtonGroup.propTypes = {
+  activeValue: PropTypes.string,
+  /** Each object in array accepts value, content, onClick, and icon, and also will pass through any extra props */
+  buttons: PropTypes.arrayOf(
+    PropTypes.shape({
+      value: PropTypes.string,
```
sfrieson-cbi: I would add an isRequired here since it's required for both UI and performance (`key` value). That'll also help us understand that it is valued higher than `content`

rey-wright: Good call. Thanks!

PR #53, opened 2019-07-10
fds icon now uses currentColor

rey-wright:
## Packages changed

- [ ] `fds-dictionary`
- [x] `fds-styles`
- [x] `fds-components`
- [ ] `fds-mui-theme`
- [x] `fds-icons`

## Description
`.fds-icon` now uses `currentColor` under the hood. I reverted some usage of currentColor in `fds-components` that is no longer needed with this change, and this class is heavily used inside `fds-icons` package as well.

## Checklist
- [X] All unit tests pass


rey-wright:
I could use some help testing this via `yarn link`... I wasn't totally able to test this... I may be missing something simple though.

packages/fds-components/src/style/interactive/buttonGroup.css
```diff
@@ -43,11 +43,6 @@
     border var(--button-group-timeout) var(--motion-easing-default);
 }
 
-.groupbtn .fds-icon {
```
akdetrick: so unfortunately we can't remove this style until `fds-components` is updated to the `2.33` of `fds-icons` in a separate PR.

rey-wright: got it. I know you explained this to me... I feel like I understood this last week, but need to wrap my head around it again. Reverting!

rey-wright:
Thanks for the review! I'll try to handle the bumping of the components library.

PR #55, opened 2019-07-16
align icons at top of Toast when there is multiline text

rey-wright:
## Packages changed

- [ ] `fds-dictionary`
- [ ] `fds-styles`
- [X] `fds-components`
- [ ] `fds-mui-theme`
- [ ] `fds-icons`

## Description
When text inside the toast is multiline the icon centering isn't a desirable outcome. Because of this, we want to top align icons, while still giving them the appearance of being centered when the text isn't multiline.

My personal interpretation of this (and the way it happens to be more straightforward in terms of implementation) is that it starts top aligning once multiline goes past 2 or so lines. If not for this, we'd need to do a lot of margin top icon shifting, and for limited benefit as this solution seems to be pretty good, and the concern for the icons only happens when there are egregious amount of lines (+3)

## Screenshots
Screen Shot 2019-07-16 at 5 56 11 PM

## Checklist
- [X] All unit tests pass


packages/fds-components/src/style/interactive/toaster.css
```diff
@@ -47,3 +53,8 @@ body {
   opacity: 0.01; /* removed from DOM on duration complete */
   transform: translateY(100px);
 }
+
+/* Targeting a nested item isn't ideal, but material may block us from targeting this directly */
+.toast-action span {
+  line-height: 1;
+}
```
rey-wright: I'm going to sort of call this out in advance for your convenience :)

Whatever line-height that is on here by default (1.75) makes the button appear "slightly" out of line with things. This is a material button so I'm not sure I have access to this span's line-height (though totally let me know).

packages/fds-components/yarn.lock
```diff
@@ -173,24 +173,24 @@ fds-dictionary@^2.29:
...
     prop-types "^15"
 
-fds-styles@^2.31:
-  version "2.31.127"
-  resolved "https://registry.yarnpkg.com/fds-styles/-/fds-styles-2.31.127.tgz#56a6f4791b2d897b933d8c366aed367336e9143d"
-  integrity sha512-QvsSNuZWeCdHDiDsnD5Wi5+cM+hlakhjdMKqrBJrXr4OYdYccJfvHCXMpK8BFkhRmbDHEgbziJdpixcSlmLUyg==
+fds-styles@^2.33:
+  version "2.33.135"
+  resolved "https://registry.yarnpkg.com/fds-styles/-/fds-styles-2.33.135.tgz#b9b4af5d1f53ff0e92c2c0f6504fef9d1530e10b"
+  integrity sha512-k+Oy1KWUCxpl8o200RjhLqPfxe542jg4/hCMvUFRVWn8Y3woIm6Sd1w08kUwMw0TkhaCM3EKhs04W7tphhKCvg==
```
rey-wright: This yarn.lock change should go away if the other active PR is merged.

packages/fds-components/src/style/interactive/toaster.css
akdetrick: this is good for now - I imagine we'll want to replace this with one of our own `Button` components when we're ready.

PR #57, opened 2019-07-17
refine buttongroup

rey-wright:
## Packages changed

- [ ] `fds-dictionary`
- [ ] `fds-styles`
- [X] `fds-components`
- [ ] `fds-mui-theme`
- [ ] `fds-icons`

## Description
https://cbinsights.atlassian.net/browse/UIP-222
Change ButtonGroup aesthetics due to updated designs

This will go out after the current 2 PR's with version bump, hence me skipping some versions. Also currently waiting on some design finalization.

## Screenshots
Screen Shot 2019-07-17 at 3 21 25 PM

## Checklist
- [X] All unit tests pass


packages/fds-components/src/style/interactive/buttonGroup.css
```diff
@@ -10,62 +16,99 @@
...
+
+/* z-index is used here to manage the border for those button
+states and their stacking order (whether they display or not).
+The border for a disabled button should appear below a border
+for a button being clicked on (:active) */
+.btngroup .groupbtn:disabled {
+  z-index: 1;
+}
+
+.btngroup .groupbtn.active {
```
rey-wright: This selector triggers some lint errors. Worth discussing? I can describe why I'm doing all of this. If it does make sense, what's the command we're using to ignore lint errors?

packages/fds-components/src/style/interactive/buttonGroup.css
```diff
@@ -1,5 +1,11 @@
 :root {
   --button-group-timeout: 250ms;
+  --button-group-focus-color: rgba(0, 51, 102, 0.32);
+  --button-group-hover-color: #DEE5EB;
+  --button-group-pressed-color: #C2CEDA;
+  --button-group-active-color: #DBE2E9;
+  --button-group-disabled-border-color: #DBE2E9;
+  --border-radius-default: 4px;
```
rey-wright: I've added a bunch of variables around some of the "questionable" category of values (colors, radius). The radius will eventually live in `fds-dictionary`. The colors will not, as it's been communicated that these are specific to the ButtonGroup.

I wonder if it's "hurtful" to do stuff like this, as if the user is inspecting the active variables in the browser, they may  be confused as to what's "global" variables, vs variables specific for a style sheet (and whether there should be some convention that differentiates "global" variables from "local" variables.

packages/fds-components/src/style/interactive/buttonGroup.css
```diff
@@ -10,62 +16,99 @@
...
 
 .groupbtn:disabled {
-  opacity: 0.5;
-  cursor: default;
+  color: var(--color-silver);
+  border-color: var(--button-group-disabled-border-color);
+}
+
+.groupbtn-icon {
+  margin-right: calc(var(--space-half) / 2);
```
rey-wright: I saw a bunch of other places where if we have a value that could be considered a division / multiplication of an existing value (given that it's still the same domain of value), to just use calc to derive the value. Let me know if this is something we'd like to continue to do.

akdetrick: Sometimes we want to subdivide existing values like this and `calc()` is a nice way to do it without introducing a new public variable. I don't want to add `space-quarter` until we find it's actually needed often

packages/fds-components/src/style/interactive/buttonGroup.css
akdetrick: What if instead of `.groupbtn.active` we create a `.groupbtn--active` class?

akdetrick:
Looks like this needs a fresh merge from `master` to bring the diff up to date and resolve some conflicts

packages/fds-components/src/style/interactive/buttonGroup.css
rey-wright: _great_ suggestion. Made that change.

packages/fds-components/src/style/interactive/buttonGroup.css
akdetrick: Anything specified on `:root` will be global. If you're concerned about devs inspecting, there are two things that I think make this okay:

1. `fds-dictionary` is the one source of truth for standard CSS vars, not the inspector
2. In chrome and firefox, custom properties are split by files the properties are defined in, not by selector:

Screen Shot 2019-07-23 at 5 19 11 PM

rey-wright: Right. I do feel like this cuts both ways, because I also wouldn't want developers to overuse these file level variables. I feel like the variables I used are smart (only things that fit into fds-dictionary theme... colors, transition timeout, etc), but a develop could see that and start coding in arbitrary spacing, flexbox stuff, etc.

PR #63, opened 2019-07-25
run yarn bootstrap automatically after yarn install

rey-wright:
https://cbinsights.atlassian.net/browse/UIP-252

## Description
Added a postinstall command so that yarn bootstrap runs directly after yarn install at the root.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "author": "CB Insights",
   "license": "MIT",
   "scripts": {
+    "postinstall": "yarn bootstrap",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -4,6 +4,7 @@
   "author": "CB Insights",
   "license": "MIT",
   "scripts": {
+    "postinstall": "yarn bootstrap",
```
akdetrick: TIL: "Certain script names are special."

PR #65, opened 2019-07-29
UIP-235: Chips!

rey-wright:
## Packages changed

- [ ] `fds-dictionary`
- [ ] `fds-styles`
- [X] `fds-components`
- [ ] `fds-mui-theme`
- [ ] `fds-icons`

## Description
https://cbinsights.atlassian.net/browse/UIP-235

Add chips component

## Screenshots
Screen Shot 2019-07-29 at 3 03 04 PM

## Checklist
- [X] All unit tests pass


packages/fds-components/src/components/interactive/Chips.jsx
```diff
@@ -0,0 +1,111 @@
...
+  hasClose: PropTypes.bool,
+  value: PropTypes.string,
+  size: PropTypes.oneOf(['small', 'medium', 'large']),
+  theme: PropTypes.oneOf(['blue', 'gray', 'outline']),
+};
+
+const Chips = ({ list, ...topRest }) => (
+  <ul className="fdsChip-ul">
+    {list.map((listItem) => (
+      <li className="fdsChip-li" key={listItem.value || listItem.label}>
```
YunfeiYan: I think it's possible to have the same value/label here. Combine the value/label with the `index` could help keep the key unique.

rey-wright: Thanks for reviewing!

So the only problem with that, is that keys aren't supposed to be based off of the index. Essentially the idea behind keys, is let's say you have 10 items. If you remove the 3rd item, you want to communicate to react to remove only that 3rd item. If you have the keys based on index, and you remove the 3rd item, the index of every item after that will update.

https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

YunfeiYan: Yes, use index is no good for performance. But if we have the same key, the list will break.

packages/fds-components/src/components/interactive/Chips.jsx
```diff
@@ -0,0 +1,115 @@
...
+  );
+};
+
+Chip.propTypes = {
+  label: PropTypes.string,
+  /**
+   * This allows for overriding of the Chip root element.
+   * Meant to accomodate  `<Link />`
+   */
+  rootEl: PropTypes.func,
```
sfrieson-cbi: I've seen the prop name `as` used before. Also `component`. `rootEl` might be slightly deceiving since it sounds like it might need to be a dom element or an already rendered react component `rootEl={<div />}` vs `component={'div'}`.

packages/fds-components/src/components/interactive/Chips.jsx
rey-wright: Totally. So for the most part, it's the consumer's responsibility to make sure to provide unique keys, not only for their own components, but for this one too.

The great thing about us controlling the mapping, is that consumers are unable to mistakenly use index as a key, so they're forced to find a better way to establish a key.

For this component, by default, it attempts to use the label as the key. Ostensibly, labels (and by proxy tags and filters) in the application "really" shouldn't be duplicated in the application (correct me if I'm wrong though)... so this should provide a correct key in 99% of cases without interaction (and protects against using index)

For the 1% case, the consumer will need to provide their own key (maybe a unique id they get from the backend... maybe they decide to generate their own uuid, etc). But yeah that's totally expected, as there's no real practical way to enforce the id's are unique on the reusable component side. The developer would see the duplicate key error in their console, and then look at the component interface to figure out how to pass a unique identifier. That make any sense at all?

The only bad thing about all of this: React actually will throw an error if you try to create a "key" prop. The name "value" could potentially be a bit more distinctive and establish a pattern across fds components?

packages/fds-components/src/components/interactive/Chips.jsx
rey-wright: yup, that makes sense. I'm down with both of those suggestions. Changing now.

rey-wright: I started with component, but felt like it didn't have enough context (sure component, but component rendered where exactly... inside? outside? replacing that base component? etc). `as` I think provides the context I'm looking for.

rey-wright:
@sfrieson-cbi thanks for the approval and the suggestion... I've made that change 👍

rey-wright:
Thanks for the reviews everyone. Going to merge this in now but let me know if you have any other thoughts.

packages/fds-components/src/components/interactive/Chips.jsx
YunfeiYan: I agree. Many thanks for the explanation!

rey-wright: absolutely, thanks for the review.

PR #75, opened 2019-08-06
119/major/monorepo refactor

akdetrick:
# Monorepo refactor

Changes project structure to publish as a single npm package.

🚨 **"Files changed" is not as scary as it looks** 🚨 
`327` files have changed, but most of the changes are renames and suppressed in the diff.

----

## Base branch
`release/4.0`

### What changed?

* removed `lerna` - all modules moved to `<root>/src/`
* updated all cross-module imports to be relative imports
* moved all build/dev scripts to `<root>/scripts`.
* moved all tooling configuration to repo root (styleguidist, mdcss, etc)
* added `assets` modules for css rollup (see "combined styles" in the table below)
* updated all documentation to reflect these changes
* updated publish pipeline in `Jenkinsfile`
	* added versioning step (yarn bumps package.json)
	* modified publish step (uses `yarn publish` directly instead of publish via `lerna`)
* `CHANGELOG` for versions < `4.0` split into a separate `CHANGELOG-v3.md`

### What didn't change?

* Bookmarks to documentation will not be broken. We can update this later if we want.
* No source code has been modified in this branch (aside from moves/renames)

### Import changes

The import paths will change with this release. This is what consumers will now use:

module           | import statement
---------------- | ----------------------------------------
dictionary       | `@cbinsights/fds/lib/dictionary/`
icons            | `@cbinsights/fds/lib/dictionary/`
components       | `@cbinsights/fds/lib/components/`
combined styles  | `@cbinsights/fds/lib/assets/all-styles.min.css`

### What does the published package look like now?

<https://unpkg.com/browse/@cbinsights/fds/>


akdetrick:
I am going to go ahead and merge this into `release/4.0` in order to unblock other features (breaking typographic updates) we want to merge into the release branch.

README.md
```diff
@@ -31,35 +24,58 @@ for details on each major and minor release.
...
----------------- | ------------------------------------------------------
-`yarn bootstrap`      | Bootstraps dependencies for packages with Lerna
-`yarn build`          | Builds unversioned `lib/` or `dist/` dirs in each package; builds all docs to `<root>/docs/`
+
+Incomplete list of yarn targets.
+
+#### Builds
+
+Some modules rely on others to be built first. It's a good idea to start with a `yarn
+build:full` before running focused builds.
```
rey-wright: I'm torn here. I know this is for performance, but I also wonder the safety of just explicitly building those packages that each package may depend on. This is one of those things where if I read it, I'm going to go "hmm... I don't know what this means, or what those packages are... I'm just going to build the full thing... or they just use the command without reading the readme, and figure out later through some Slack conversation that there were instructions to using these commands.

src/components/interactive/ButtonGroup.md
```diff
@@ -3,10 +3,10 @@ Hardcoded example without any local state
 ```js
 import ButtonGroup  from './ButtonGroup';
 
-import FeedIcon from 'fds-icons/lib/react/FeedIcon';
-import TableIcon from 'fds-icons/lib/react/TableIcon';
-import MarketMapIcon from 'fds-icons/lib/react/MarketMapIcon';
-import StarEmptyIcon from 'fds-icons/lib/react/StarEmptyIcon';
+import FeedIcon from '../../../lib/icons/react/FeedIcon';
```
rey-wright: I'm not the biggest fan of relative paths like this... I'm generally only a fan of relative paths things that are located together (so this markdown file and the component are together, so it makes sense to use relative paths) but for things completely different, I find it introduces a lot of finickiness, doesn't do a great job communicating (developer importing this depending on how deep their folder structure is in their given use case will have to use trial and error each time drilling up), and could be more consistent.

There should be a way to refer to these things "absolutely" IMO, e.g.

```
import Blah from '/lib/blah/blah'
```

That way, anything someone needs to consume something inside this package, it's a uniform path, similar to what we had before when they were actual packages.

There's a way to set this up with either babel or webpack tweaks if I recall correctly.

PR #82, opened 2019-08-13
119/r4/rm material ui

akdetrick:
## Description
- adds `MuiButtonShim` and `MuiAvatarShim` as basic components to fill in for `material-ui` lib components
- refactor `MenuButton`, `AvatarRow`, and `Toast` to replace `material-ui` components with our shims
- remove `material-ui` from dependencies

## Checklist
- [x] All unit tests pass


src/components/media/MuiAvatarShim.jsx
```diff
@@ -0,0 +1,61 @@
...
+ * This shim replaces material-ui Avatar to cover usage in FDS.
+ * Renders a basic avatar.
+ */
+const MuiAvatarShim = ({ name, src, size, className, style }) => {
+  const classNames = [
+    'fdsAvatar',
+    'alignChild--center--center',
+    `media--${size}`,
+    className,
+  ].join(' ');
```
rey-wright: This is just a shim, so not sure how closely I should be looking at these things 😄but wondering why you opted to go with array.join here vs classnames. I feel like the shim itself is temporary but assuming it wasn't, I feel like using the array with class join will encourage people to keep using it, even when they do need a classname in this list to be conditionally rendered, and then they'll just array.push it in.

src/components/style/interactive/menuButton.css
```diff
@@ -1,12 +1,8 @@
...
-
 .menuButton {
-  padding: 4px !important;
-  color: var(--font-color-primary) !important;
-  text-transform: none !important;
-  min-height: var(--media-l) !important;
+  padding: 4px;
+  color: var(--font-color-primary);
+  text-transform: none;
+  min-height: var(--media-l);
```
rey-wright: 🔥

src/components/style/interactive/muiButtonShim.css
```diff
@@ -0,0 +1,33 @@
...
+  border-width: 0;
+  padding: calc(var(--space-half) / 2) var(--space-half) !important;
+  border-radius: var(--border-radius-default);
+  cursor: pointer;
+  color: var(--font-color-secondary);
+  fill: var(--font-color-secondary);
+  font-weight: var(--font-weight-bold);
+  text-transform: uppercase;
+  margin: 0;
+}
```
rey-wright: Is the `fdsButton` name going to conflict with our upcoming button component stuff? I wonder if we should start namespacing these different buttons now to avoid naming clashes & semantic confusion (so `fdsMuiButton`, `fdsFooButton`, etc)

src/components/media/MuiAvatarShim.jsx
akdetrick: I didn't think `classnames` was necessary here, but you're right, we don't want this pattern to spread. I'll update it

src/components/style/interactive/muiButtonShim.css
akdetrick: Upcoming button components will replace the shim, so I'm not worried about collisions here

PR #84, opened 2019-08-14
Bringback prepush

rey-wright:
## Packages changed
N/A

## Description
I think the pre-push command got removed in the monorepo feature (I assume it was either because it was making development a pain for that big feature specifically, or the general behavior of the prepush hook isn't desirable). In any case, definitely want to bring it back for yarn test, and we can sort of decide if we'd like to remove the rest of the command.

## Screenshots
N/A

## Checklist
N/A


rey-wright:
@akdetrick I'm going to just go with yarn test for now, and we can circle back to this later on.

PR #86, opened 2019-08-16
NotPerfect.md

rey-wright:
## Description
- I mostly wanted to quickly jot this out before bedtime. 
- This absolutely shouldn't be the final form of this message (nor should this be off alone in it's own file) but the core theme of the idea is there. "Don't wait for the perfect component api to contribute".
- mostly want to see if we align on the underlying message, if you have any ideas on how we can best convey this, but happy to keep refining it into something more presentable.

## Checklist
- [ ] All unit tests pass

### DIFF
--- a/NotPerfect.md
+++ b/NotPerfect.md
@@ -1,0 +1,7 @@
+Unless the perfect api interface is "obvious", optimize for fds component creation, rather than getting the api right.
+
+Developers may have the inclination that they cannot contribute new component that they're working on to this new repo. This may be because they're afraid they haven't built a component that accommodates for all future use cases.
+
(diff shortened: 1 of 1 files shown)


NotPerfect.md
```diff
@@ -0,0 +1,7 @@
+Unless the perfect api interface is "obvious", optimize for fds component creation, rather than getting the api right.
+
+Developers may have the inclination that they cannot contribute new component that they're working on to this new repo. This may be because they're afraid they haven't built a component that accommodates for all future use cases.
```
akdetrick: This all makes a lot of good points; to build the most future-proof components there is a tradeoff in the number of components you can build and the number of contributors that we can gain.

There should be a balance here, and I agree that we are currently leaning too heavily on the side of optimizing for the right API rather than for optimizing for component creation.

One big thing we can do to shift the balance is to make the right design principles more "obvious" by documenting conventions and patterns that all reusable components should follow. I will prioritize those docs this week, as I think it could go a long way toward addressing this concern.

It would be great if we could open up FDS to more contributors. The 4.0 release was a good step in this direction, but we can do more.

rey-wright:
@akdetrick I feel your documentation you've written adequately (https://github.com/cbinsights/form-design-system/pull/88) encapsulates a lot of what was communicated here 👍closing.

PR #88, opened 2019-08-19
Add contribution guidelines

akdetrick:
Adds guidelines to repo.

## Reviewing
I recommend just going here to read through because rendered markdown is a little nicer to read:
https://github.com/cbinsights/form-design-system/blob/119/docs/UIP-211_contributing-guidelines/CONTRIBUTING.md

## Revisions
If there are any guidelines in here you are opposed to, let's find the right solution and update them before merging.

## Future plans
This is by no means complete, but it gets us started.


contributing/release-and-publish.md
```diff
@@ -0,0 +1,29 @@
...
+
+
+## Publishing docs
+All FDS documentation is published to [Github Pages](https://pages.github.com/).
+
+The publishing source is the `docs/` directory in repo root. Documentation is regenerated
+on every full build and **we commit generated documentation**.
+
+Generated documentation is suppressed in pull requests diffs by the `linguist-generated`
+gitattribute.
```
rey-wright: hmm... never heard of that before!

We can potentially add more explanation here: "We generate documentation locally which we commit. We generally don't want to see this documentation in our pull request diffs as it complicates things. Our documentation uses a `linguist-generated` gitattribute which suppresses the files being showed in the diff: https://help.github.com/en/articles/customizing-how-changed-files-appear-on-github

We do this because we commit the generated files for now, which means we don't want to use gitignore" and then add a link that explains more about that attribute: https://help.github.com/en/articles/customizing-how-changed-files-appear-on-github

contributing/principles.md
```diff
@@ -0,0 +1,90 @@
...
+It is unlikely that a custom JS solution will perform as reliably as something implemented
+natively in a browser.
+
+For example, `position: sticky` should be preferred over a JS solution to fixing content
+during scroll.
+
+### JS-in-JS, CSS-in-CSS
+In FDS, we have opted for keeping styling as close to native web standards as possible.
+We do not use any CSS-in-JS libraries. We do not use CSS preprocessors. CSS is written
+based on the latest CSS standards (with help from PostCSS for polyfills and vendor sugar).
```
rey-wright: I think PostCSS in a sense IS a preprocessor (polyfills, vendor). You could technically enable Sass like features pretty easily with it.

rey-wright: I'm currently working on a "CSS in JS" article that _could_ be a candidate for a good blog post. If so, we could drop a link to that right in here.

contributing/styling.md
```diff
@@ -0,0 +1,85 @@
+[&larr; Table of Contents](../CONTRIBUTING.md)
+
+# Styling Guidelines
+🎨 How to CSS in FDS
+
+## Styling components
+
+Prefer CSS custom properties over hard coded values. The [style dictionary](https://cbinsights.github.io/form-design-system/fds-dictionary/) in fds lists available custom properties.
+
```
rey-wright: Just a thought: a developer could easily interpret this as "oh this just means to make variables at the top of my file and reuse them in my declarations" rather than "try to go find existing variables for this", "if there's a close value, try to see if you can get the value changed to an existing value", etc. I can totally see people re-coding variables that already exist at the top of their specific component css file because they misunderstood what was the intention here.

contributing/naming-conventions.md
```diff
@@ -0,0 +1,150 @@
...
+```diff
+-<Menubutton />
++<MenuButton />
+```
+
+Prefer simple component names. Namespacing is unnecessary.
+
+```diff
+-<FdsButton />
++<Button />
```
rey-wright: 👍

contributing/naming-conventions.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+
+## Components
+
+Use `PascalCase` for component names.
+
+```diff
+-<Menubutton />
++<MenuButton />
```
rey-wright: 👍

contributing/naming-conventions.md
```diff
@@ -0,0 +1,150 @@
...
+
+```css
+/* parent-child */
+.base-child {}
+
+/* modifier */
+.base--modifier {}
+
+/* child modifier */
+.base-child--modifier {}
```
rey-wright: 🔥

contributing/naming-conventions.md
```diff
@@ -0,0 +1,150 @@
...
+
+#### Conditional
+Some classes may apply their styles conditionally based on browser support, viewport size,
+or other contextual condition. We indicate this with a prefix and underscore at the
+beginning of the class name.
+
+- **`.atMedium_`**`margin--all` (adds margin to all sides when the viewport reaches the
+- medium breakpoint)
+- **`.atLarge_`**`display--block` (sets display to block when the viewport reaches the
+- large breakpoint)
```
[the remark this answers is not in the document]
rey-wright: I know you marked conditional classes as not existing yet above... it may be worth not calling out these utilities until they exist... I can totally see some developer trying to use this.

contributing/principles.md
```diff
@@ -0,0 +1,90 @@
...
+- Components that rely on Redux state
+- Components that are tied to a specific application or feature
+- Components that are not "small and sharp" (see below)
+
+## Alignment with design
+When it comes to standard UI components, Engineering and Design should speak the same
+language. In practice, this means:
+
+- Publicly exposed components in FDS should be named based on how design refers to the
+component in their sketch library
```
rey-wright: I 100% agree but this being the case, we'll want to start doubling down to make sure design doesn't reflect business use cases (There's been at least one component so far which we've discussed that calls out specific business cases. )

contributing/principles.md
```diff
@@ -0,0 +1,90 @@
...
+designing a generic component by following the ["Worse is Better"](https://en.wikipedia.org/wiki/Worse_is_better) method.
+
+Here are some key points of this method as it applies to FDS:
+- Simplicity is the most important consideration in a design.
+- Components should be consistent, **but not at the sacrifice of simplicity**.
+- Components should cover as many use cases as possible, but **completeness is not as important as simplicity**
+
+This means that a component only needs to support the absolute minimum functionality before shipping.
+
+**Favor software designs that can be iterated on as needed instead of optimizing for complexity up-front.**
```
rey-wright: 🔥

contributing/styling.md
```diff
@@ -0,0 +1,85 @@
...
+```
+/* "base" includes all default styles. Variant classes only change what is necessary */
+<Component className="base base--variant base--anotherVariant" />
+```
+
+**Avoid binary class switching**. Use an additive class pattern instead.
+
+```diff
+-className={isSelected ? 'menu-item-selected' : 'menu-item-normal'}
++className=cx('menu-item', { 'menu-item--selected': isSelected })
```
rey-wright: Is it worth calling out the actual classnames (or classcat) package? Essentially it's not easily possible to do this additive approach without these packages.

contributing/editor.md
```diff
@@ -0,0 +1,29 @@
...
+## Eslint
+This project uses [eslint](https://eslint.org/) to lint JS files.
+
+- Download an eslint plugin for your editor here: <https://eslint.org/docs/user-guide/integrations#editors>
+
+
+## Stylelint
+This project uses [stylelint](https://stylelint.io/) to lint CSS files.
+
+- Download a stylelint plugin for your editor here: <https://stylelint.io/user-guide/complementary-tools#editor-plugins>
```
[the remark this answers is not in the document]
akdetrick: I have drifted from VSCode, but would welcome changing this file. Maybe we could break it down by popular editors instead?

```
## VSCode setup
- eslint plugin link
- stylelint plugin link
- prettier plugin link

## Sublimetext
- ""
```

contributing/styling.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Styling typography
+
+**Avoid setting font properties in CSS**. Prefer utility classes provided by FDS:
+
+CSS rule                                   | utility class
+------------------------------------------ | --------------------
+🚫 `font-size: var(--font-size-xl)`        | ✅ `.fontSize--xl`
+🚫 `font-weight: var(--font-weight-bold)`  | ✅ `.fontWeight--bold`
+🚫 `color: var(--font-color-secondary)`    | ✅ `.color--secondary`
```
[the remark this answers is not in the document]
akdetrick: In my experience, even with a class-only approach to styling, I haven't seen any elements with more than 6 classes.

Utility classes don't cover _everything_, so I expect we'll still see engineers writing new CSS for additional styling. The key point I want to make here is that engineers should _prefer_ the utility classes, especially for font properties.

---

One exception might be if you see the same few classes being used together a lot, which means a new abstraction may be necessary.

If we pretend for a minute that `type--data` doesn't exist, imagine if we saw these 3 classes being repeated a lot in the code:

```
className="fontSize--small fontWeight--normal color--secondary"
```

That would indicate the design has a concept for this style of type that is not reflected in code. The solution here would be to work with design to name the thing and create a new class, like `type--data`.

contributing/release-and-publish.md
akdetrick: Added some clarity in this section

contributing/principles.md
akdetrick: Editing this to clear up what I actually mean by "preprocessors"

contributing/styling.md
akdetrick: Going to make this statement explicit about preferring the vars provided by FDS

contributing/principles.md
akdetrick: This sounds like more of a thing for us to keep in mind as we move forward. Design so far has been basing common components off of feature designs which definitely include business.

Is there anything we should add to these docs regarding that though?

contributing/styling.md
akdetrick: i opted for the familiar library here - we just have to remember to search `./contributing` for usage when we switch to classcat

contributing/principles.md
rey-wright: hmmm... yeah this is tough. I think this comes back to a lot of this: https://github.com/cbinsights/form-design-system/pull/88#pullrequestreview-277219170

Essentially, do we want to spell out that developers should push back against business logic being inserted into components, or push more on the design side so that there isn't.

Me and Asaf have been having some really good discussions as of late that may inform some of this as well (we should talk about this in our 1:1 probably). I think we can easily just say that we'll come back to this.

contributing/styling.md
rey-wright: right. So what I mean here is like... let's assume classcat doesn't exist. I wonder if developers even know what `cx` is.

contributing/editor.md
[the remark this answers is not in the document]
rey-wright: I think the plugins mostly look great. There's a setting developers will generally want, after they install the Prettier plugin, that triggers the formatting to update on save, which doesn't come out of the box. I can add this at a later date though (but to be fair, this is heavily based on how many of our front end developers actually use VSCode).

contributing/styling.md
```diff
@@ -0,0 +1,85 @@
...
+
+**Avoid setting `line-height`.** The utility classes provided by FDS set an appropriate
+line height for type based on font size.
```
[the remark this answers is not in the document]
akdetrick: I think we should avoid setting line height in CSS entirely if we can, and when we do, we should use a standard line height value (e.g. `var(--font-lineHeight-smallText`).

Avoiding `line-height` in general might help us move away from anti-patterns I've seen in our code:

- using a unit for line height values, like `17px`
- using line height to control element hight (`height` or `padding` should be used instead)

contributing/styling.md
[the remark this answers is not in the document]
rey-wright: I agree to a degree. I actually think a lot of the explanation you just gave would be right at home in this PR somewhere 🙂

contributing/principles.md
rey-wright: I think it goes back to the point I had at the very very top: https://github.com/cbinsights/form-design-system/pull/88#pullrequestreview-277219170

I think this is something worth covering in our 1:1 though, for us to think about and we can always come back if we so choose.

akdetrick: Sounds to me like this is a wider concern than what contributing guidelines for FDS can cover. Let's discuss how best to encourage pushing back from feature teams

contributing/styling.md
akdetrick: good point - going with plain JS for this example.

contributing/styling.md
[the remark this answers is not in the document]
akdetrick: Added an example to styling guidelines

PR #97, opened 2019-09-03
Add chip on close

rey-wright:
## Description
This PR re-instates the following PR: https://github.com/cbinsights/form-design-system/pull/92

Essentially, our feature designs call for the chip, and the close icon, to be separately interact-able. Even if they didn't, it may be ambiguous to the user that clicking on the tag itself would / could be responsible for closing the tag.

## Screenshots
N/A

## Checklist
- [x] All unit tests pass


rey-wright:
@lzhang-cbi Since you worked on this, we've killed some of the broader line-height declarations, which is why I didn't re-instate those, and Bootstrap as a whole should be gone this coming Monday (September 9th).

src/components/interactive/Chip.jsx
```diff
@@ -37,7 +47,7 @@ Chip.propTypes = {
   Link: PropTypes.func,
   isActive: PropTypes.bool,
   subtitle: PropTypes.string,
-  hasClose: PropTypes.bool,
+  onClose: PropTypes.func,
```
akdetrick: Let's jsdoc all these prop types (the generated props table will pick up jsdoc descriptions)

PR #103, opened 2019-09-10
122/base styles/add radius helper

akdetrick:
## Description

- Adds `rounded--<direction>` helper classes for border radius

## Screenshots
Screen Shot 2019-09-09 at 6 16 00 PM

## Checklist
- [x] All unit tests pass


src/base-styles/helperClasses/border/radius.css
```diff
@@ -0,0 +1,38 @@
...
+  border-radius: 0 var(--border-radius-default) var(--border-radius-default) 0;
+}
+
+.rounded--bottom {
+  border-radius: 0 0 var(--border-radius-default) var(--border-radius-default);
+}
+
+.rounded--left {
+  border-radius: var(--border-radius-default) 0 0 var(--border-radius-default);
+}
```
rey-wright: The way you have it, you wouldn't be able to combine `rounded--right` with `rounded--left` because you're zero'ing out the other corners in each declaration. Is this the intention (and if so, should we call this out in documentation)?

rey-wright: oh. nvm I see what you're doing :)

PR #106, opened 2019-09-11
Adding Button

rey-wright:
**Not urgent to code review, but please take a look if you're interested!**

## Description
Adding a standardized button.

## Screenshots
Screen Shot 2019-09-11 at 6 53 21 PM

## Checklist
- [x] All unit tests pass


src/components/style/interactive/button.css
```diff
@@ -0,0 +1,90 @@
...
+.fdsButton--loading::before {
+  content: "";
+  box-sizing: border-box;
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  width: 16px;
+  height: 16px;
+  margin-top: -8px;
+  margin-left: -8px;
```
rey-wright: I'm not "sure" I like the idea of using variables here, given that these values always need to be half of the width and height of this element (so if the variables ever change in the future, this loader would break.... (#resilientCSS))

src/components/style/interactive/button.css
```diff
@@ -0,0 +1,83 @@
+.fdsButton {
+  --button-blue: 0, 102, 153;
+  --button-red: 229, 50, 62;
+  border: 0 none;
+  padding: var(--space-half) var(--space-default);
+  border-radius: var(--border-radius-default);
+  font-weight: 500;
+  text-transform: uppercase;
```
rey-wright: I could technically use like 4 utility classes here... but seems a bit excessive.

akdetrick: We'll want to revisit the font weight - we currently do not include a 500 weight on cbi-site.

Once that's fixed, we can introduce a `--fontWeight-medium` var.

src/components/interactive/Button.jsx
```diff
@@ -0,0 +1,78 @@
...
+        >
+          <Icon size="xs" />
+        </div>
+      )}
+    </Element>
+  );
+};
+
+Button.defaultProps = {
+  theme: 'primary--blue',
```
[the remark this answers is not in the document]
akdetrick: *tl;dr* - we should talk to design about this a bit more. We should resolve these questions:

- Supporting every color/type combination will provide a more flexible long-term solution. How does design feel about this? The tradeoff of not doing this means that we can never design a button that is any color other than blue or red (unless we refactor the API to what I am suggesting)
- When we go back to replace buttons in existing features, should we force all of them to blue or red?

----

I guess it would look more like...

```js
{
  type: PropTypes.oneOf('contained', 'outline', 'ghost'),
  color: PropTypes.oneOf('blue', 'red', ... ),
}
```

I have a few reasons for bringing this up:
- being able to independently declare type and color is scalable while maintaining a clear distinction between color and type
- the `theme` prop values introduce a new concept, which is a string that looks like a class but isn't actually a class
- the `theme` prop is behaving like two separate props in one, which is probably an indication it should be two props
- if we support some combinations and not others, this could get confusing to devs. If i see `primary--blue` and `ghost--red`, i might expect there would be a `ghost--blue`, but there's not. I'd have to look up docs every time I use a button to see what's valid.

I understand why you set it up this way though - there are limitations in color/type combinations that are dictated by the design.

On taking a closer look at designs, I suspect this won't scale. If we ever want a purple button for any reason, we're forcing developers to make a custom button, which eschews all of the resilience and best practices we build into the standard component.

Sometimes guard rails are nice, but if they're too restrictive, it encourages "off-roading" if a designer or engineer doesn't like where the component is telling them where to go.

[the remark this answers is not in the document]
akdetrick: With all that said, our release habits mean that we can in fact make a breaking change to this later to support a different API if we need it. Let's discuss this first with design and make a call from there.

src/components/style/interactive/button.css
```diff
@@ -0,0 +1,192 @@
...
+.fdsButton--outlined:active,
+.fdsButton--ghost:active {
+  background-color: var(--button-secondary-pressed);
+  color: var(--button-secondary-pressed-color);
+}
+
+.fdsButton--outlined.fdsButton--isDestructive:hover,
+.fdsButton--ghost.fdsButton--isDestructive:hover {
+  background-color: var(--button-secondary-destructive-hover);
+}
```
rey-wright: So you might be thinking here: why don't you do this: 

```
.fdsButton--isDesctructive:matches(.fdsButton--outlined, .fdsButton--ghost) {}
```

And the answer is... it doesn't work straightforward? Something is happening here where what should be the correct order and specificity isn't being preserved.

src/components/interactive/Button.jsx
```diff
@@ -0,0 +1,100 @@
...
```
akdetrick: Need to change this default prop to reflect new theming options

rey-wright: Gooood catch.

src/components/style/interactive/button.css
rey-wright: ^ incorrect @rey-wright . I've updated with a working example. You're welcome.

PR #110, opened 2019-09-20
Add Wrapper prop to GroupButton

sfrieson-cbi:
## Description
Adds a prop `wrapper` the the `GroupButton` to support features like tooltip than need to directly wrap their child.
```js
<Tooltip>
  <GroupButton />
</Tooltip>
```

Also adds private props to `GroupButton` to continue to support the first and last buttons' styling differences.

I could not think of a good example to put into Storybook that wouldn't promote misuse of the prop.

Made some other small cleanups to fix tests and remove some console errors.

## Checklist
- [x] All unit tests pass


src/components/interactive/ButtonGroup.jsx
```diff
@@ -5,20 +5,41 @@ import combine from '../../util/combine';
...
-    {buttons.map(({ value, content, key, Icon, onClick, label, ...restBtn }) => (
-      <GroupButton
-        key={key || label}
-        onClick={onChange ? combine(() => onChange(label || value), onClick) : onClick}
-        label={label}
-        Icon={Icon}
-        {...restBtn}
-      />
-    ))}
+    {/* `isFirstButton` and `isLastButton` are removed to guard against them being passed in. */}
```
rey-wright: I don't know if this is really necessary to explicitly called out. There's a bunch of other stuff here that's also been removed so that it's not passed into rest props.

src/components/interactive/buttonGroup.test.js
```diff
@@ -10,8 +10,8 @@ const Icon = () => (
 )
 
 const buttons = [
-  { value: 'Feed', icon: Icon, isActive: true },
-  { value: 'Table', icon: Icon, onClick: () => {}, disabled: true },
+  { label: 'Feed', Icon, isActive: true },
+  { label: 'Table', Icon, onClick: () => {}, disabled: true },
```
rey-wright: uhhh.... whoops? 😄

src/components/interactive/GroupButton.jsx
```diff
@@ -38,6 +55,19 @@ GroupButton.propTypes = {
...
+   * `wrapper` must render the children prop passed into it
+   * to render the GroupButton.
+   */
+  Wrapper: PropTypes.elementType,
+
+  /** For internal use: Assigns the first button styling. */
+  isFirstButton: PropTypes.bool,
+
+  /** For internal use: Assigns the last button styling. */
+  isLastButton: PropTypes.bool,
```
rey-wright: Just to talk this through:

So GroupButton already by "proxy"?... is already an internal button. People "can" use it, but I don't really see that being a thing, at least at the current moment. Let's say they did have a reason to use it... maybe they would have a reason to use these props? Essentially, I just wonder the usefulness of calling out that these are "internal props" when this component is private to begin with and doesn't even show up in Storybook.

src/components/interactive/groupButton.test.js
```diff
@@ -5,11 +5,14 @@ import GroupButton from './GroupButton';
...
-)
+);
+const Wrapper = (props) => (
+  <div {...props} data-test="snapshot-wrapper" />
+);
 
 describe('GroupButton component', () => {
   it('matches snapshot (set all props)', () => {
-    expect(shallow(<GroupButton value='Feed' icon={Icon} isActive={true} />)).toMatchSnapshot();
+    expect(shallow(<GroupButton value='Feed' Icon={Icon} isActive={true} Wrapper={Wrapper} />)).toMatchSnapshot();
```
rey-wright: Good catches.

src/components/interactive/ButtonGroup.jsx
sfrieson-cbi: I called this out because they're removed _and_ not passed in or used otherwise. The others are. Just wanted to be clear that they weren't accidentally left off.

src/components/interactive/GroupButton.jsx
sfrieson-cbi: Hmm... I see what you're saying. If this component ever were public, these would no longer be private. Since it is currently private these props don't matter to them. Will remove.

src/components/interactive/ButtonGroup.jsx
rey-wright: Talked through offline. Makes perfect sense to me. We would never want the consumer to be able to define isFirstButton on any ol random button 👍

PR #113, opened 2019-09-23
Button stories background

rey-wright:
## Description
It's important for our designers (and us) to know how these buttons render on different background colors. We actually are waiting to be bitten because I know for a fact some of our buttons interaction states are supposed to use rgba / transparency, but they don't. The perfect way to address this (outside of a multitude of other benefits) is to provide a way for designers to go in to see the transparency applied now, which means providing a knob to toggle the background color (I can then make a hands off Monday ticket to supply us with colors after this)

_(well... after Storybook knobs are fixed, but that's a different story)_

I also fixed up some proptype errors in the Button stories example.

## Screenshots
Screen Shot 2019-09-23 at 11 06 12 AM

## Checklist
- [x] All unit tests pass


src/components/interactive/button.stories.js
```diff
@@ -18,33 +18,40 @@ storiesOf('Interactive/Button', module)
...
-              outlined: 'outlined',
-              ghost: 'ghost',
-            },
-            'blue',
-            { display: 'inline-radio' }
-          )}
+        <div
+          className={boolean('Toggle Background', false) ? 'padding--all inverted' : ''}
+          style={
+            boolean('Toggle Background', false) ? { background: 'var(--color-navy)' } : {}
```
akdetrick: good use of knob 🎛

PR #121, opened 2019-09-30
Added font-weight: 500 to Button, added 500 Roboto font, removed 300 Roboto font

rey-wright:
## Description
Added font-weight: 500 to Button, added 500 Roboto font, removed 300 Roboto font

## Screenshots
Screen Shot 2019-09-30 at 10 20 37 AM

## Checklist
- [x] All unit tests pass


src/components/style/interactive/button.css
```diff
@@ -28,6 +28,7 @@
   --button-secondary-destructive-pressed: #ECD6D9;
   border: 0 none;
   letter-spacing: 1px;
+  font-weight: 500;
```
akdetrick: Should we add this to `dictionary`, or wait until `cbi-site` actually supports 500 weight?

rey-wright: @akdetrick yeah it's an interesting question. My thought is that we "shouldn't", only because it would only be used internally for FDS. The reason for that, is Asaf expressed interest in only using this for Buttons and pretty much nothing else, and I'd like all of our reusable Buttons to be defined in FDS... so essentially, unless a developer is making a new button in FDS (not likely), or editing an existing button in FDS (not likely), they would never need to use 500 weight directly. But let me know what you think.

PR #123, opened 2019-09-30
123/component/uip 381 section

akdetrick:
## Description
- Add `bgColor--<color>` background helpers
- Add `border--<side>` helpers
- Add `--double` variants for spacing helpers
- Add `Section` component, tests, and stories
- Add `storybook.js` utils for convenience writing stories

## Screenshots
Screen Shot 2019-09-30 at 2 20 39 PM
Screen Shot 2019-09-30 at 2 21 04 PM
Screen Shot 2019-09-30 at 2 21 17 PM
Screen Shot 2019-09-30 at 2 21 23 PM

## Checklist
- [x] All unit tests pass


src/base-styles/helperClasses/border/sides.css
```diff
@@ -0,0 +1,54 @@
...
+  border: var(--standardBorder);
+}
+
+.inverted:matches(
+.border--top,
+.border--right,
+.border--bottom,
+.border--left,
+.border--all,) {
+  border-color: var(--font-color-hintInverted);
```
rey-wright: Should be be using `font-color` to style `border-color`?

src/components/layout/Section.jsx
```diff
@@ -0,0 +1,102 @@
...
+    getPaddingClasses('h', hPadding),
+    getPaddingClasses('v', vPadding),
+    {
+      inverted: VALID_BG_INVERTED.includes(bgColor),
+      [getBorderClasses(border)]: Boolean(border),
+    }
+  );
+  const styles = {
+    backgroundClip: 'border-box',
+  };
```
rey-wright: Should we make a CSS Module for this, just so no one gets any funny ideas? (or, reusable class)

src/components/layout/Section.jsx
```diff
@@ -0,0 +1,102 @@
...
+    <div style={styles} className={classNames}>
+      {children}
+    </div>
+  );
+};
+
+Section.defaultProps = {
+  hPadding: 'default',
+  vPadding: 'default',
+  bgColor: 'white',
```
rey-wright: Would we support transparent background colors with this? (and if so, would white or transparent operate as the default)?

src/components/layout/Section.jsx
```diff
@@ -0,0 +1,102 @@
...
+
+export const VALID_PADDING = ['default', 'double', 'half', 'none'];
+
+/**
+ * @param {String} axis 'h' or 'v'
+ * @param {String} amount
+ * @returns {String} padding utility classes
+ */
+export const getPaddingClasses = (axis, amount) => {
+  const sides = axis === 'h' ? ['left', 'right'] : ['top', 'bottom'];
```
akdetrick: @rey-wright any strong feelings about `h`/`v` versus `x`/`y`?

rey-wright: No strong strong feelings... but I do think x & y as concepts are a lot more apparent what they're doing. Like if you saw just x & y in a vacuum, and asked them what these things meant, they might say "oh it represents an axis or a chart) or something like that, whereas `h` and `v` themselves aren't as apparent. I will say that it's slightly weird that we're spelling out each one but these (though not too weird).

So yeah personal preference in order, would be to use x/y, then spell them out with the full name, and then use h/v... but not super strong feelings any way.

src/components/layout/Section.jsx
```diff
@@ -0,0 +1,102 @@
...
+      classes = `border--${direction}`;
+  }
+  return classes;
+};
+
+/**
+ * @param {Object} props react props
+ * @returns {ReactElement}
+ */
+const Section = ({ hPadding, vPadding, bgColor, border, children }) => {
```
rey-wright: This is something that might be interesting to talk about. What components should use rest props, vs which ones shouldn't, and why. I'm personally not sure if I'd want to do the thing where a developer has to ask for data attribute or whatever first, for us to add it... I feel like we should either support rest props, or support nothing, as we're doing here.

src/base-styles/helperClasses/border/sides.css
akdetrick: In this case, yes - we're taking advantage of the alpha transperency of hint text for inverted color sections

src/components/layout/Section.jsx
akdetrick: yeah, can't hurt

src/components/layout/Section.jsx
akdetrick: I'm thinking that `Section` should always set a background color, as background color is part of its responsibility. If we support a transparent version, the background color is less contained to the section component and relies on parent components to manage backgrounds

src/components/layout/Section.jsx
akdetrick: agreed, going to go with `x`/`y`; those names sit better with me because at least they're more familiar

src/components/layout/Section.jsx
akdetrick: `data-test` is the best reason I can think of with supporting rest props, and I don't think `Section` is the right place to put that attribute anyway.

I'm of the opinion that the webdriver should be hooking into specific content inside layout components, not the layout components themselves.

src/components/layout/Section.jsx
rey-wright: Hmmm that doesn't entirely make sense to me. So like... you can also set border in Section but it's not required. I also do feel like in a lot of cases, you'll have one white thing (let's just say a `Card`) and assuming design doesn't change color, you're just trying to divide up the sections... so would you force the developer to, in every section inside that white card, redefine it to be white... and it would make it harder to go back and change the color, and it would also make any sort of color change on interaction for the card non-doable, as you'll have a bunch of white boxes on top. 

We should chat about this just so we're aligned and I get where you're coming from.

src/components/layout/Section.jsx
rey-wright: Talked offline. Layout components shouldn't get rest props (and this is a layout component).

src/components/layout/Section.jsx
rey-wright: Talked offline. This actually makes a lot more sense to me now, and forces the concern of display to be always on section, rather than fall through, and thereby forces our colors to always be accurate (you can't put a section on green and have it display as green).

PR #125, opened 2019-10-04
UIP-391: <Dialog /> and <Prompt />

rey-wright:
## Screenshots
Screen Shot 2019-10-04 at 10 19 40 AM
Screen Shot 2019-10-04 at 10 19 34 AM

## Description

https://cbinsights.atlassian.net/browse/UIP-391

- Dialog: the base component for all Modals,
- Prompt: a more specific confirm / deny component

- This uses a similar mounting strategy to Toaster: the actual component itself can't be unmounted due to CSS Transition, however the Modal itself is completely wiped from the DOM when `isOpen` is falsy.
- The idea is that Dialog represents the base "modal" for all other modals to be built with. Developers can use Dialog for their general use case... However if they have a simple message that has a primary button and a secondary button to confirm / deny, that would be a better use case for `Prompt`.
- Don't worry about design. We're going to be polishing this up, but design has already reviewed and is 90% on board here, just may need some general padding tweaks we can follow up with in another PR.
- This uses our new `<Section />` component. We'll be talking about this in frontend group probably in the near future.
- This adds two new libraries: `no-scroll`, and `focus-trap-react`. The latter is for accessibility purposes, the former is for disabling scrolling when the modal appears.

Things to do in the future:

- We want to give a hook to developers to specify something in the content that should be focused first? Maybe?
- Need to add the logic where focus is returned to the element that brought focus to it. This is another fairly easy addition on top of dialog.

## Checklist
- [x] All unit tests pass


src/components/modals/dialog.test.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import Dialog from './Dialog'
+
+describe('Dialog component', () => {
```
tonysight: I wonder if instead of just snapshot testing, we should test some of the important behavior like if `onDismiss` will in fact be called or the scrolling.

src/components/modals/Prompt.jsx
```diff
@@ -0,0 +1,48 @@
...
+import Flex from '../layout/Flex';
+import FlexItem from '../layout/FlexItem';
+
+const Prompt = (props) => (
+  <Dialog
+    content={
+      <React.Fragment>
+        <div id="a11y-dialog-desc">{props.desc}</div>
+        <div className="margin--top--double">
+          <div className="">
```
tonysight: empty `className`

rey-wright: Good catch 👍updated

src/components/modals/dialog.test.js
rey-wright: True.

Me and Adam are going to possibly talk about a more robust testing strategy next week potentially (I really don't like snapshots) but in the meantime, I can add those tests.

src/components/modals/Dialog.jsx
```diff
@@ -0,0 +1,120 @@
...
+                        )}
+                      </Flex>
+                    </Section>
+                  </div>
+                </React.Fragment>
+              )}
+              <div className="dialog-content">
+                <Section>{props.content}</Section>
+              </div>
+              {props.footerContent && (
```
tonysight: Looking at `Prompt`, if the bottom buttons are passed in `props.content` which then toggles `props.isOpen` outside of this component, then when will `props.footerContent` be used?

rey-wright: Yeah that's a very good question, with a complicated answer:

footerContent is for modals that are larger, so that... when the modal scrolls, the buttons aren't hidden at the bottom (if that happens, we'd make developers construct that modal manually by hand). The reason buttons aren't in the footer now, is because the footer is gray.

That being said, we have an action item to talk with design regarding this. We may always want buttons in the footer, but we don't necessary always want them on a gray background, so we may require some sort of theme prop to control whether the footer will be gray or white, and then the buttons will appear the same as they do now, except they'll actually be in the footer.

akdetrick:
Adding "do not merge" label to give us an opportunity to make any tweaks we may need after design review on Monday.

src/components/modals/Dialog.jsx
```diff
@@ -0,0 +1,120 @@
...
+};
+
+Dialog.propTypes = {
+  /** Controls the role of the modal */
+  role: PropTypes.oneOf(['dialog', 'alertdialog']),
+  /** Controls whether the modal (and overlay) are shown or not */
+  isOpen: PropTypes.bool,
+  /** Controls whether the modal can be cleared passively (esc key, clicking
+   * backdrop, showing x on modal, clicking on modal x to close) */
+  canDismiss: PropTypes.bool,
```
tonysight: Instead of having `props.canDismiss`, would it be better if it's calculated based on if `props onDismiss` is passed a function?

akdetrick: `onDismiss` is an optional callback - our thinking was that an engineer shouldn't have to pass a noop to trigger a dismissible dialog

src/components/modals/Dialog.jsx
```diff
@@ -0,0 +1,120 @@
...
+
+const Dialog = (props) => {
+  const handleClick = () => {
+    if (props.canDismiss) {
+      props.onDismiss();
+    }
+  };
+
+  const handleKeyDown = (e) => {
+    if (props.canDismiss && e.key === 'Escape') {
```
akdetrick: does babel lend any help with this in the transpiled component? `e.key` may need to be polyfilled

src/components/modals/Dialog.jsx
```diff
@@ -0,0 +1,120 @@
...
+    // is shown or not
+    if (props.isOpen) noScroll.on();
+    noScroll.off();
+  }, [props.isOpen]);
+
+  return ReactDOM.createPortal(
+    <CSSTransition timeout={200} in={props.isOpen} classNames="dialog" unmountOnExit>
+      <React.Fragment>
+        <FocusTrap>
+          <div>
```
akdetrick: is this extra wrapping dev necessary?

src/components/modals/Dialog.jsx
rey-wright: so from what I researched, e.key is supported even up until IE9.... it's just that IE has non-standard key values... so you'll have people doing stuff like `e.key === "ArrowDown" || e.key === "Down`. All the polyfill things I saw where slightly underwhelming, and from reading the docs, `Escape` seemed supported on IE as well, but haven't had a chance to check. Let me get Browserstack up and running to see if this works on IE11.

src/components/modals/Dialog.jsx
rey-wright: Yeah good question. The only thing that would determine if we can do this, is if the consumer fully controls the dismiss otherwise... so on modals like prompt... let's just imagine FDS controlled the buttons for dismissing... even though we don't let the user click the x mark, we'd still want the ability to dismiss the modal.

That being said... I'm not "sure" that's a thing. You may be right that we can derive this just from the function.

src/components/modals/dialog.test.js
akdetrick: :whynotboth: 🙃

src/components/style/interactive/dialog.css
```diff
@@ -0,0 +1,68 @@
...
+}
+
+.dialog-overlay {
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  position: fixed;
+  /* leaving the zero addition here for now */
+  z-index: calc(var(--zindex-modal) + 0);
```
akdetrick: we have a `--zindex-shim` for this - do you think that could work well enough?

src/components/style/interactive/dialog.css
```diff
@@ -0,0 +1,68 @@
...
+  overflow-y: auto;
+}
+
+.dialog-icon {
+  cursor: pointer;
+  border: 0 none;
+  min-height: auto;
+}
+
+/* .dialog-divider {
```
akdetrick: ✂️

src/components/modals/Dialog.jsx
tonysight: If `props.isOpen` determines if a modal is open and that is being controlled outside of this component, to dismiss this, they will need to pass a `props.onDismiss` to toggle `props.isOpen`. If not, I suppose `handleClick` and `handleKeyDown` will need to check if `props.onDismiss` is passed a function?

src/components/modals/Dialog.jsx
rey-wright: Yeah. I couldn't tell you why, but `<FocusTrap />` expects a single element, and it throws if you try to pass it a `<React.Fragment />`

src/components/style/interactive/dialog.css
rey-wright: I think so. I'll adjust.

src/components/modals/Dialog.jsx
rey-wright: 😄 this is difficult to explain.

I think the best course of action is to move forward with your solution, and we can re-examine this at another time. What I'm thinking of isn't needed now and could in fact be a pre-optimization.

(and yeah handleClick & handleKeyDown will need to check onDismiss, which is fine).

src/components/modals/dialog.test.js
rey-wright: I don't dislike many things, but snapshot tests are one of them 🙂That being said, I've added both.

src/components/modals/dialog.test.js
```diff
@@ -0,0 +1,51 @@
...
+
+  // Dialog component uses a library that uses window.scroll underneath the hood
+  // so we just need to add this here to avoid errors
+  // eslint-disable-next-line no-undef 
+  window.scroll = () => {}
+
+  it('dismisses modal when close icon is clicked', () => {
```
akdetrick: we should also cover the ESC close in tests

src/components/modals/dialog.stories.js
```diff
@@ -0,0 +1,30 @@
...
+                // eslint-disable-next-line no-console
+                console.log('close modal');
+              }
+            : undefined
+        }
+        title={text('title', 'Modal title here')}
+      />
+    ),
+    { notes: { markdown: README } }
+  );
```
akdetrick: this could use another story or two to demonstrate:

- actions in the footer
- enough content in the modal to trigger scrolling behavior

src/components/modals/prompt.test.js
```diff
@@ -0,0 +1,33 @@
...
+import { shallow } from 'enzyme';
+import Prompt from './Prompt'
+import Button from '../interactive/Button'
+
+describe('Prompt component', () => {
+
+  // Dialog component uses a library that uses window.scroll underneath the hood
+  // so we just need to add this here to avoid errors
+  // eslint-disable-next-line no-undef
+  window.scroll = () => { }
```
akdetrick: might be a good idea to define this in `scripts/node/setupTests.js` instead

src/components/modals/Dialog.jsx
```diff
@@ -0,0 +1,142 @@
...
+    }
+  };
+
+  const [isOverflowing, setIsOverflowing] = useState(false);
+
+  const handleResize = () => {
+    rafSchd(setIsOverflowing(isElementOverflowing(contentEl)));
+  };
+
+  useLayoutEffect(() => {
```
akdetrick: not a blocker, but i think this could use regular ol' `useEffect` because the resize listener doesn't need to do any DOM API manipulation

akdetrick: not a blocker, but i think this could use regular ol' `useEffect` because the resize listener doesn't need to do any DOM API manipulation

PR #126, opened 2019-10-04
124/component/uip 399 inlineblocklist

akdetrick:
## Description
- Adds `InlineBlockList`
- Removes margins from `Chip`

## Screenshots
Screen Shot 2019-10-04 at 11 47 08 AM
Screen Shot 2019-10-04 at 11 47 14 AM
Screen Shot 2019-10-04 at 11 47 25 AM
Screen Shot 2019-10-04 at 11 47 33 AM
Screen Shot 2019-10-04 at 11 47 38 AM

## Checklist
- [x] All unit tests pass


src/components/style/layout/inlineBlockList.css
```diff
@@ -0,0 +1,25 @@
+.fdsInlineBlockList {
+  --inlineBlockList-gutter: var(--space-half);
```
akdetrick: oops, we don't need this anymore

src/components/layout/InlineBlockList.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import cx from 'classnames';
+import PropTypes from 'prop-types';
+
+/**
+ * @param {Object} props react props
+ * @returns {ReactElement}
+ */
+const InlineBlockList = ({ items, separator }) => {
+  const classNames = cx('fdsInlineBlockList', {
```
rey-wright: I wonder if we want to give users the ability to pass in ` ` here (a space), as they can sort of adjust the size of the spacing indirectly in that way.

src/components/layout/inlineBlockList.stories.js
```diff
@@ -0,0 +1,84 @@
...
+    ),
+    { notes: { markdown: README } }
+  )
+  .add('Chips list', () => (
+    <InlineBlockList
+      items={[
+        <Chip size="lg" onClose={noop} label="Horses" theme="outline" />,
+        <Chip size="lg" onClose={noop} label="Wombats" theme="outline" />,
+        <Chip size="lg" onClose={noop} label="Seals" theme="outline" />,
+        <Chip size="lg" onClose={noop} label="Dogs" theme="outline" />,
```
rey-wright: are the `noop`s being passed in here because some Chip functionality fails without them?

rey-wright: are the `noop`s being passed in here because some Chip functionality fails without them?

src/components/layout/InlineBlockList.jsx
rey-wright: I wonder if we want to give users the ability to pass in `SPACE` here (an empty string with a space), as they can sort of adjust the size of the spacing indirectly in that way.

rey-wright:
One more thing: it would be cool if we could also add this to the Chip variations story

src/components/layout/inlineBlockList.stories.js
akdetrick: `onClose` is needed to show the close icon

rey-wright: got it 👍

src/components/layout/InlineBlockList.jsx
akdetrick: yeah, technically someone could do this: `<InlineBlockList separator="          "`

If we see anybody doing this, it's a sign we should probably have a second prop to control spacing:

`spacing: PropTypes.oneOf(['quarter', 'half', 'default', 'double'])`

PR #134, opened 2019-10-11
UIP-413: added small button variant

rey-wright:
## Description
Request from Nectar. Added a size prop, made the current variant medium (`m`), and added the small (`s`) variant. Asaf signed off as well.

## Screenshots
Screen Shot 2019-10-11 at 11 19 49 AM

## Checklist
- [x] All unit tests pass


src/components/style/interactive/button.css
```diff
@@ -35,6 +35,11 @@
   background-color: transparent;
   transition: background-color var(--button-group-timeout) var(--motion-easing-default),
     box-shadow var(--button-group-timeout) var(--motion-easing-default);
+  min-height: auto;
+}
+
+.fdsButton--s {
+  padding: calc(6px - 1px) calc(var(--space-half) - 1px);
```
akdetrick: a `calc()` on two hard coded values could just be `5px` instead. not a big deal though

rey-wright: Yeah. I wanted to try to preserve the context of what I was doing here. So the spacing values are actually 6 x 8, but I'm knocking off a pixel because of the transparent border. I figure if someone else comes in here and adds updated spacing, they'll update the 6px and leave the 1px alone?

PR #139, opened 2019-10-15
124/v5/uip 417 add adrs

akdetrick:
## Description

Adds `adr` aka "Architectural Decision Records". An ADR is a record of a significant architectural decision (e.g. reorganize the component hierarchy).

I have chosen the MADR (markdown ADR) format for FDS. 

## Context

There are some discussions happening around documentation. We can give ADRs a test drive in our project and report our findings to the rest of engineering. 

My choice of ADRs comes down to the principle of keeping docs as close to the code as possible. This type of documentation also fills the gap of _why_ a certain decision was made.

## Checklist
- [x] All unit tests pass


reywright:
Cool. I couldn't find any sort of linting hook. That would be cool to sort of verify if people are writing these in the correct format.

akdetrick:
here is a "frequently imagined questions" from me to address some general questions about what/why/how ADRs fit into our workflow:

> When do we write them?
I don't know. I'm introducing ADRs so we can find out in practice what works. I think we'd only add an ADR for a "significant" architectural design change. We can be the best judge of what that actually means.

> When do we read them?
They're for other people or our future selves to understand _why_ certain decisions were made.
It's also useful to scan existing ADRs in the case of creating a decision the supersedes a prior ADR.
Otherwise, we don't have to pay much attention to them.

> How long do we try this out?
I want to give it a few months. We can see how it's going near the middle or end of Q1. We can gather our findings of what did or did not go well with ADRs and either abandon them or apply our lessons to other projects at CBI.

PR #142, opened 2019-10-18
Pass the event to onClose and preventDefault

jjniczCB:
## Description

preventDefault needed if the parent is an `<a/>`
Also passing the event, if the consumer want's to use it.

## Checklist
- [] Docs have been rebuilt (`yarn build:full`)
- [x] All unit tests pass
- [] Snapshots have been updated


akdetrick:
seems like more of a bugfix than a minor version, so i don't think we need to update the version/changelog

PR #154, opened 2019-10-31
Update every single package

rey-wright:
## Description
Ran `yarn upgrade`. We're now fully up to date (`yarn upgrade-interactive` renders no packages that can be updated)

I split this up from the React / Babel / Storybook updates, just in case anything goes awry updating, and then we can isolate off which is the actual culprit better (but hopefully it doesn't get to this).

## Screenshots
N/A

## Checklist
- [x] All unit tests pass


rey-wright:
Note: whenever this is approved, I'm going to re-run `yarn upgrade`.

PR #155, opened 2019-10-31
removed lodash (replaced with snake-case library)

rey-wright:
## Description
Removed `lodash`.

I downloaded the first widely used snake case library I found (`snake_case`) which actually fixes a small bug we had. View the screenshots for what changed when I ran `yarn build:dictionary` when I started using this new function.

## Screenshots
Screen Shot 2019-10-31 at 2 46 53 PM
Screen Shot 2019-10-31 at 2 56 55 PM

These are not what the variables actually "are" I believe, so they're currently displaying wrong.

## Checklist
- [x] All unit tests pass


akdetrick:
If the variable names change, I'm cool with rolling this up into `release/v5` and marking the breaking change. Worth it for removing lodash.

rey-wright:
Screen Shot 2019-10-31 at 4 50 44 PM

@akdetrick So according to the docs, this is actually the way they're supposed to be, and I just verified that.

`FONT_SIZE_2_XL` is wrong. `FONT_SIZE_2XL` is right. Essentially, not only are we getting rid of lodash, but we're fixing an error in our documentation, hence this is not actually a breaking change, as `FONT_SIZE_2_XL` doesn't exist nor work, but `FONT_SIZE_2XL` does.

akdetrick:
`_2XL` is the right way to document the var in this branch, but `master` has `_2_XL`:
https://unpkg.com/browse/@cbinsights/fds@4.35.249/lib/dictionary/js/styleConstants.js

The documentation we publish from `master` reflects `_2_XL`:
https://cbinsights.github.io/form-design-system/fds-dictionary/

Looks like the main issue is that it's incorrect in the CHANGELOG

rey-wright:
@akdetrick wow I could have sworn I console logged 2XL... good catch! Let me look at this again bright and early Monday.

rey-wright:
Unless you have any reservations, I'll remake this change for 5.0. Upon research, it "does" seem like there is debate on how snake case should convert, especially when numbers are involved. I don't believe `cbi-site` is even using it, and me personally, I prefer `2XL` vs `2_XL`... so I'll make that new PR for now.

PR #157, opened 2019-10-31
V5: UIP-450: dogfooding buttons for Toast

rey-wright:
## Description
Our Toast component uses non standard buttons. We have `IconButton` and `Button` now so we can just use those.

This "could" go into v4, but the only catch is that I needed to change the interface around. CountdownButton accepts "text" as the icon, rather than an actual icon, so I had to rip up that interface, and have it accept an FDS icon.

Now... "technically" we're not using CountdownButton in `cbi-site` directly, so this wouldn't induce a breaking change there. That being said, if someone was using this in a branch we don't know about, or if the public were using this, then it would.

Last thought: we could maybe get away with putting this in v4 if we figured out a way to keep the prior interface intact, whilst supporting this new one. Since no one is banging down our doors for this, and I don't have the brainpower right now to figure that out, I figured I'd just make this a v5 change.

## Screenshots
Screen Shot 2019-10-31 at 7 03 45 PM

## Checklist
- [x] All unit tests pass


rey-wright:
@akdetrick this branch technically contains a breaking change 😄do you want to just slate this for v6? We could have a quicker turn around for v6 if we wanted too.

akdetrick:
Adding "do not merge" label for now. We may want to close this PR and open a new one when we're working on v6

akdetrick:
Reopening; this can go into v6

akdetrick:
@rey-wright Made some changes to shore this up with `release/6.0`. Could use a quick review from you to make sure I didn't step on any of your changes (looks good to me when I run storybook).

- Update `CHANGELOG`
- Remove `MuiButtonShim`

rey-wright:
Thanks for shoring this up @akdetrick ... looks good to me.

PR #165, opened 2019-11-05
move source order

akdetrick:
## Description

It doesn't look like we need to important any other helpers until someone reports an issue. I did however, change the cascade a bit (e.g. `flush` classes should always override spacing helpers)

## Checklist
- [x] All unit tests pass


rey-wright:
Sweet

PR #168, opened 2019-11-12
UIP-478: add ability to pass href or Link into IconButton

rey-wright:
## Description
We originally did not want to include this, but the site is actually making use of it. This is one of those situations where it would not be worth it to (right now) mandate that someone change the designs in any case where you have an href on an IconButton (a good example is that wacky profile page fixed header). So while we can promote that people don't do this, we should support this so that we don't have to keep accommodating the older icons

## Screenshots
N/A

## Checklist
- [x] All unit tests pass


akdetrick:
Been giving `Link` support more thought...

What if, instead of giving this component extra responsibilities just for handling one specific wrapper type, we rely on composition instead:

**Composing a link in a consumer application**
```
<Link ...>
    <IconButton as="div" ... />
</Link>
```
Because buttons inside anchors are not valid, we'd have to support an `as` prop that changes the DOM element type.

The advantage to this approach is that it leads to less "special case" logic in each individual component, and it makes the component smaller and sharper. It's also less duplicated logic between components that support a `Link`.

What do you think? If we want to go the composition route, we could queue up some v6 changes to existing components that support a `Link` prop.

akdetrick:
**Update**

`<Component Link={Link} to="url" />` was the decision from our discussion. The tradeoffs are worth it.

rey-wright:
@sfrieson-cbi you mind giving me a review on this one? Glad to discuss it whenever, if you have the bandwidth.

PR #173, opened 2019-11-15
UIP-485: dogfood z-index css properties in FDS

rey-wright:
## Description
We'd like to use the z-index css properties in cbi-site, but we're not even using them in FDS internally. Let's set a good example 🙂

## Screenshots
N/A

## Checklist
- [x] All unit tests pass


src/components/style/interactive/buttonGroup.css
```diff
@@ -9,35 +9,35 @@
...
+  z-index: var(--zindex-main);
   position: relative;
 }
 
 .btngroup .groupbtn {
   margin-left: -1px;
   position: relative;
   border-radius: 0;
-  z-index: 2;
+  z-index: calc(var(--zindex-main) + 1);
```
YunfeiYan: In cbi-site we may see something like: `z-index: calc(var(--zindex-main) + 99999);` Lol

rey-wright: 🙂Technically you're not allowed to add more than 9. We're planning to go fix all the z-index values up in cbi-site, and then hopefully we'll catch anyone trying to do that in code review (or add some sort of lint rule).

PR #174, opened 2019-11-15
UIP-443: added docblocks & mdx

rey-wright:
## Description
The folder structure I think with docblocks wants things to be separated out by section, but I'm not sure I can separate the older components not using docblocks out by section. As a result, Interactive is a section, but everything else is in an "other" folder right now. I'm fine with this (again, assuming we want to take an intermediary approach, and tackle it component by component) but we can re-examine.

## Screenshots
Screen Shot 2019-11-15 at 4 27 50 PM

## Checklist
- [x] All unit tests pass


rey-wright:
There are some.... slight bugs with the modal docs. There still could be a good argument for pushing this now (we can just call that a bug with modals we'll deal with later, it creates a good reproducible solution for us to post around for them to fix, etc). Developers can always go to that specific markdown file for modals in the meantime.

rey-wright:
I've now fixed all the bugs in modals (I did this by upgrading to Storybook beta, and also disabling docs mode for all dialog and prompt modals)

PR #182, opened 2019-11-21
add DecoratedInput component

akdetrick:
## Description
**Branched from (and merging to) `release/6.0` to get form style updates**

- Adds `DecoratedInput` component.
- right-aligns number-ish inputs

## Screenshots
Screen Shot 2019-11-21 at 5 40 30 PM

## Checklist
- [x] All unit tests pass


rey-wright:
Just because I have to ask 🙂:

Screen Shot 2019-11-22 at 10 20 10 AM

How closely do you see this work being correlated with componentization of the above pattern? Essentially, this is controlling some kind of token, positioned either at the beginning, or end of the input. I'm assuming we'd want to also use this for icons, and then the question becomes, would the above just be a ghost style or something.

It could be that this does become apart of this work, but we can think about it more fully at a later time / date, or these could be completely different constructs. Just wanted to make sure to raise it.

src/components/forms/decoratedInput.stories.js
```diff
@@ -0,0 +1,53 @@
...
+      </React.Fragment>
+    ),
+    { notes: { markdown: README } }
+  )
+  .add(
+    'Variations',
+    () => (
+      <Flex direction="column">
+        <FlexItem>
+          <DecoratedInput key="both" before="£" after="GBP">
```
rey-wright: what is this `key` prop for?

akdetrick:
Icons will not be in side labels - for inputs with an icon on the inside, something else should be responsible for composing that. A `SearchInput` for example, would absolute an icon inside the input and set input padding.

src/components/forms/decoratedInput.stories.js
akdetrick: oops - it used to be in an array; will remove

PR #184, opened 2019-11-25
UIP-424: Fix modal centering and appearance

rey-wright:
## Description
Modal has both had a hard time appearing, and being centered. This should finally, at last fix that.

## Screenshots
N/A

## Checklist
- [x] All unit tests pass


src/components/style/modals/dialog.css
```diff
@@ -17,14 +21,21 @@
 }
 
 .dialog-overlay {
+  z-index: var(--zindex-shim);
+  background: var(--color-charcoal);
+  opacity: 0.6;
+}
+
+.dialogfullScreen {
```
akdetrick: why not apply the `z-index` and positioning rules directly to `dialog-overlay` and `dialog-wrapper`?

rey-wright: @akdetrick the fix is in

PR #190, opened 2019-12-04
add IconInput

akdetrick:
## Description
Adds `IconInput`

## Screenshots
Screen Shot 2019-12-04 at 11 30 16 AM
Screen Shot 2019-12-04 at 11 30 27 AM

## Checklist
- [x] All unit tests pass


rey-wright:
My only question here, is why this element wouldn't spit out an input itself, rather than requiring you to wrap an input. 

I sort of get why you did this (composition) but I just wonder.... if this should always be used with an input and nothing else, if there is advantage here this not outputting an input element itself (and using spread props).

src/components/style/forms/iconInput.css
```diff
@@ -0,0 +1,29 @@
...
+.fdsIconInput--hasEndIcon input {
+  padding-right: var(--space-double);
+}
+
+.fdsIconInput-icon {
+  position: absolute;
+  top: 1px;
+  bottom: 1px;
+  display: flex;
+  align-items: center;
```
rey-wright: In my own components, I've been using an extra div to utilize `alignChild--center--center`. This is more of a lesson to me, but it seems you're ok curving that class and spelling it out directly here?

src/components/forms/iconInput.test.js
```diff
@@ -0,0 +1,46 @@
...
+      <IconInput
+        IconLeft={StarFilledIcon}
+      >
+        <input type="text" />
+      </IconInput>
+    ).find('.fdsIconInput');
+
+    expect(rootEl.hasClass('fdsIconInput--hasStartIcon')).toBe(true);
+    expect(rootEl.hasClass('fdsIconInput--hasEndIcon')).toBe(false);
+  });
```
rey-wright: I like itttttt

src/components/forms/IconInput.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+IconInput.propTypes = {
+  /** An `input` element child */
+  children: PropTypes.node,
+
+  /**
+   * FDS Icon _reference_
+   * `IconLeft={StarFilledIcon}`
+   */
+  IconLeft: PropTypes.func,
```
rey-wright: 👍 good explanations

src/components/forms/IconInput.jsx
```diff
@@ -0,0 +1,51 @@
...
+import FDS from '../../../lib/dictionary/js/styleConstants';
+
+/**
+ * @param {Object} props react props
+ * @returns {ReactElement}
+ */
+const IconInput = ({ IconLeft, IconRight, children }) => {
+  const classNames = cx('fdsIconInput', {
+    'fdsIconInput--hasStartIcon': Boolean(IconLeft),
+    'fdsIconInput--hasEndIcon': Boolean(IconRight),
```
rey-wright: do these need `Boolean` or were you just trying to be more transparent that these values aren't boolean to begin with

src/components/style/forms/iconInput.css
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.fdsIconInput-icon--start {
```
akdetrick: i guess these could be utility classes

src/components/style/forms/iconInput.css
akdetrick: good call - I'll use the utility

src/components/forms/IconInput.jsx
akdetrick: mostly for clarity - it's technically not necessary

akdetrick:
Reasons for using `children` to pass the input:

- Same interface as `DecoratedIcon`
- Avoids putting input-related props on the `IconInput` component (keeps `IconInput` itself smaller and sharper - only takes props related to the icons)
- `<input {...inputProps}` is easier to reason about than `<NotTheInput {...inputProps}` - there's more transparency into which props are going where

rey-wright:
@akdetrick that makes sense... though I'd like to make sure I fully understand, only because if we extrapolate that argument, I'm pretty sure we'd have a bunch more existing FDS components that debatably should be abstracted into multiple components if this is the principle we're abiding by.

PR #193, opened 2019-12-10
UIP-510: <Avatar /> 

rey-wright:
## Description
Not this Avatar: https://www.imdb.com/title/tt0499549/

Supports intials (1-2), images (that fallback to the initials), small and medium size, as well as a bunch of colors.

## Screenshots
Screen Shot 2019-12-10 at 3 03 09 PM

## Checklist
- [x] All unit tests pass


src/components/interactive/Avatar.jsx
```diff
@@ -0,0 +1,70 @@
...
+   * ```
+   * import { Link } from 'react-router'
+   * Link={Link}
+   * ```
+   */
+  Link: PropTypes.func,
+  /** Controls color of button */
+  bgColor: PropTypes.oneOf(VALID_BG_COLORS),
+  /** Controls the size of the button */
+  size: PropTypes.oneOf(['s', 'm']),
```
tonysight: Should we add support for a size smaller than `'s'` too? @akdetrick and I were talking about potentially having this replace the initial avatar in `MemberEdit` dropdown options: https://github.com/cbinsights/cbi-tables/blob/master/src/edit/MemberEdit/Img.jsx

akdetrick: +1 on supporting `xs` here

rey-wright: Screen Shot 2019-12-10 at 5 04 14 PM

Asaf is hoping we can use the medium size ones here.

src/components/media/MuiAvatarShim.jsx
```diff
@@ -13,7 +13,7 @@ const VALID_SIZES = ['s', 'm', 'l', 'xl'];
  */
 const MuiAvatarShim = ({ name, src, size, className, style }) => {
   const classNames = cx(
-    'fdsAvatar',
+    'fdsMUIAvatar',
```
akdetrick: I would vote to remove this component with the introduction of `Avatar` - replace usage of `MuiAvatarShim` with `Avatar`

src/components/style/media/avatar.css
```diff
@@ -0,0 +1,39 @@
+.fdsAvatar {
+  font-size: var(--font-size-xs);
+  user-select: none;
+  transition: background-color var(--button-timeout) var(--motion-easing-default),
```
akdetrick: V tiny nitpick - we don't need to transition `background-color` on this

src/components/interactive/Avatar.jsx
```diff
@@ -0,0 +1,70 @@
...
+  'lightGray',
+  'orange',
+  'charcoal',
+  'navy',
+  'aqua',
+  'gray',
+  'red',
+];
+
+const INVERTED_BG_COLORS = ['gray', 'charcoal', 'navy', 'orange'];
```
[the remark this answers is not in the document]
rey-wright: Instead of whitelisting colors to invert, I basically changed this to blacklist things that shouldn't be inverted:

`const NON_INVERTED_BG_COLORS = ['white', 'haze', 'lightGray'];`

src/components/style/media/avatar.css
rey-wright: Adjusting! Going to remove transition altogether now that we don't have the focus style.

src/components/interactive/Avatar.jsx
```diff
@@ -0,0 +1,70 @@
...
+  return (
+    <Element
+      {...rest}
+      title={name}
+      className={cx('fdsAvatar', `bgColor--${bgColor}`, `fdsAvatar--${size}`, {
+        'color--white': INVERTED_BG_COLORS.includes(bgColor),
+      })}
+    >
+      {imgUrl && (
+        <span className="fdsAvatar-img" style={{ backgroundImage: `url(${imgUrl})` }} />
```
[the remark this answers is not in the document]
rey-wright: @akdetrick makes sense. One thing though: `name` is a required field. So ignoring case #2: my first test has the name being passed (case #1), and no imgUrl passed (case #4), and the 2nd test has imgUrl being passed (case #3). I could add a separate test for case #4 though.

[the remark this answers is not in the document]
akdetrick: oh! I missed the fact it's required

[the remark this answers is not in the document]
rey-wright: Screen Shot 2019-12-11 at 2 23 59 PM

This work?

src/components/interactive/Avatar.jsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import baseElement from '../../util/baseElement';
+
+export const VALID_BG_COLORS = [
```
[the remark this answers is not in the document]
rey-wright: @tonysight the color of those should be purple 👍

src/components/media/MuiAvatarShim.jsx
rey-wright: totally. So my only question here: MuiAvatarShim supports large and extra large, which Avatar does not right now. I currently don't see any usage of `AvatarRow` in cbi-site. 

We could totally either:

-  deprecate the `large` and `xl` sizes that AvatarRow supports (this would require a version bump)
- Add support for `large` and `extra large` in Avatar... but the question there would be... do we need those.

src/components/interactive/Avatar.jsx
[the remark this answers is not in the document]
akdetrick: that would cover it

src/components/media/MuiAvatarShim.jsx
akdetrick: good call - removing `l` and `xl` would require a version bump, but deprecating them would not.

We could map unsupported values in `AvatarRow` to supported avatar sizes for now

```
// in `AvatarRow`
const sizeMap = {
  xs: 's',
  s: 's',
  m: 'm',
  l: 'm',
  xl: 'm',
};
```

src/components/interactive/Avatar.jsx
```diff
@@ -0,0 +1,57 @@
...
+  'orange',
+  'charcoal',
+  'navy',
+  'aqua',
+  'gray',
+  'red',
+  'purple',
+];
+
+const NON_INVERTED_BG_COLORS = ['white', 'haze', 'lightGray'];
```
akdetrick: Not a blocker, but a minor stylistic tip.

Some colors belong to both lists, which can be prone to error if someone modifies one of these lists later without modifying the other. One potential improvement is to make each color belong only to one list:

```
export const VALID_LIGHT_COLORS = [
	'white',
	'haze',
	'lightGray',
];

export const VALID_DARK_COLORS = [
	'gray',
	'charcoal',
	'orange',
	'red',
	'purple',
	'aqua',
	'navy',
];
```

For prop types, you can spread them:

```
	bgColor: PropTypes.oneOf([
		...VALID_LIGHT_COLORS,
		...VALID_DARK_COLORS,
	]),
```

And you can check the dark colors list directly in your `cx()`, similar to what you're already doing:

```
{ 'inverted': VALID_DARK_COLORS.includes(bgColor) }
```

src/components/interactive/Avatar.jsx
```diff
@@ -0,0 +1,57 @@
...
+  size: 'm',
+};
+
+Avatar.propTypes = {
+  /** Controls color of button */
+  bgColor: PropTypes.oneOf(VALID_BG_COLORS),
+  /** Controls the size of the button */
+  size: PropTypes.oneOf(['s', 'm']),
+  /** Control initials displayed, and also used for title accessibility attribute (uses 2 words max for initials) */
+  name: PropTypes.string.isRequired,
```
tonysight: Just remembered, previously full name wasn't required when adding a guest to share the collection. You can just provide the guest's email address. In that case, should we not make the name required?

akdetrick: I would vote on keeping this required because it allows us to set a `title` attribute.

src/components/interactive/Avatar.jsx
```diff
@@ -0,0 +1,57 @@
...
+const grabInitials = (str) =>
+  str
+    .split(' ')
+    .splice(0, 2)
+    .reduce((prev, curr) => prev + curr.charAt(0), '');
+
+const Avatar = ({ bgColor, imgUrl, size, name, ...rest }) => (
+  <div
+    {...rest}
+    aria-title={name}
```
akdetrick: should we set `role="image"`?

rey-wright:
thanks for the code reviews! @tonysight @akdetrick

rey-wright:
@akdetrick I added the `role=img` thing, good call, and I shored up some of your other requests.

rey-wright:
Going to leave `AvatarRow` alone for now.

PR #197, opened 2019-12-18
UIP-495: added border-focus helper class

rey-wright:
## Description

Added a standard class (`.border-focus`) for adding what has to come to be our standard focus style for FDS interactive elements.

- This was an existing ticket I created for this, but decided to get this done a bit sooner (thanks @jjniczCB for the good question regarding if there was a class like this that already existed).

- I went back and added this class to every interactive element with these focus styles, added this class, and purged the custom focus styles / variables that were no longer needed.
- I also added this to `Chip`, the only interactive element that does not have this.
- I'm not super attached to the name of the class or the documentation naming. Happy to take any input there.

## Screenshots
Screen Shot 2019-12-18 at 3 18 49 PM
 
## Checklist
- [x] All unit tests pass


src/components/interactive/StackedButton.jsx
```diff
@@ -10,10 +10,17 @@ const StackedButton = ({ Icon, Link, label, disabled, isActive, hasCaret, ...res
...
-      className={cx('fdsStackedButton', 'rounded--all', 'fdsStackedButton--ghost', {
-        'fdsStackedButton--disabled': disabled,
-        'fdsStackedButton--active': isActive,
-      })}
+      className={cx(
+        'fdsStackedButton',
+        'rounded--all',
+        'fdsStackedButton--ghost',
+        'border-focus--notrsn',
+        'trsn-default',
```
akdetrick: in this case, what is setting the `transition-proprety`?

src/base-styles/helperClasses/border/focus.css
```diff
@@ -0,0 +1,24 @@
...
+
+Adds standard focus style to elements
+
+class                    | description
+------------------------ | -------------------------------------------------------
+`border--focus`          | Adds standard box shadow focus style
+`border--focus--notrsn`  | Adds standard box shadow focus style without transition
+*/
+
+.border-focus:focus {
```
akdetrick: name mismatch with docs - should be double hyphen (`border--focus`, `border--focus--notrsn`)

src/base-styles/helperClasses/animations/transitions.css
```diff
@@ -0,0 +1,18 @@
...
+title: Transitions
+name: transitions
+section: Animation Helpers
+---
+
+Adds standard transitions
+
+class          | description
+-------------- | ---------------------------------------------------------
+`trsn-default` | transitions all properties to default duration and timing
```
akdetrick: Should use double hyphen `subject--verb` / `concept--modification` in name.

I know I gave a truncated example for the "transition" in this class name, but I'm wondering if we can trade off brevity for clarity in the name here:
- `trans--type`
- `transition--type`

I'd vote for spelling out `transition` for clarity. The time it takes to type the letters is trivial compared to the time it takes to understand what you're reading.

src/base-styles/helperClasses/animations/transitions.css
```diff
@@ -0,0 +1,18 @@
...
+*/
+
+.trsn-default {
+  transition-property: all;
```
akdetrick: `transition-property` has an initial value of `all`; we don't need to set it here

src/base-styles/helperClasses/animations/transitions.css
```diff
@@ -0,0 +1,18 @@
...
+---
+
+Adds standard transitions
+
+class          | description
+-------------- | ---------------------------------------------------------
+`trsn-default` | transitions all properties to default duration and timing
+*/
+
+.trsn-default {
```
akdetrick: for transitions, we may want to use more descriptive names like...

- `.transition--short`
- `.transition--medium--easeOut`
- `.transition--flash`
etc

src/base-styles/helperClasses/animations/transitions.css
rey-wright: I'm ok with `transition--type`

src/base-styles/helperClasses/animations/transitions.css
rey-wright: right. Would we want to do that now? I sort of added it this way, as I feel like we don't really have an idea what these different permutations should be.. and what should be short, medium, long, and if those would even be used. So essentially, I'm thinking we could just use `transition--default` and then build upon that if a designer specifically asks for a different timeout. I think maybe putting something about this pattern in some comments might be a nice idea though, when we need to cross that bridge.

src/components/interactive/StackedButton.jsx
rey-wright: Right. So again it really depends on how we want to sort of taper a more streamlined version for people. We can definitely have classes that just set properties and delays, but this class is more the recognition that in most of our cases right now (the case being all the interactive components), we can easily get by with the same transition delay, easing function, and property.

src/base-styles/helperClasses/animations/transitions.css
rey-wright: It doesn't work for some reason if I don't put it (Chrome)

src/components/interactive/StackedButton.jsx
akdetrick: I gave this some more thought and realized that there's always an option for an element to override `transition-property`, so I think you approach in setting that property in `transition--default` makes sense.

Let's say we want to use a standard transition, but only want to apply it to a single CSS property:

```html
<!-- transitions all properties by default -->
<div class="transition--default myClass" />
```

```css
.myClass {
        /* override to limit transition to background color only */
	transition-property: background-color;
}
```

src/base-styles/helperClasses/animations/transitions.css
akdetrick: Would be nice to make the decision later, but changing class names would be a breaking change. Just double checking this to make sure our naming scheme here scales.

Does the concept of a "default" transition make sense across the whole system? That any property being transitioned would fit a default delay, duration, and timing function?

If so, let's keep `transition--default`. If not, let's name it something like `transition--shortEase`.

akdetrick: Here's my concern with `default` as a name:

class       | what it looks like it does
----------- | ------------------------------------
`--flashLinear` | super fast linear transition
`--bounceInFast`    | bounces in fast
`--easeOutSlow`    | for slowly fading an element out or something
`--default`    | ???

To be fair, we solved this problem with things like `fontSize` by making `fontSize--default` do the same thing as `fontSize--m`.

We can always map the `--default` variant to a more descriptive name if we need to scale these classes.

src/base-styles/helperClasses/border/focus.css
```diff
@@ -0,0 +1,24 @@
...
+`border--focus--notransition`  | Adds standard box shadow focus style without transition
+*/
+
+.border--focus:focus {
+  outline: 0;
+  box-shadow: 0 0 0 2px #B7D7E5;
+  transition: box-shadow 250ms var(--motion-easing-default);
+}
+
+.border--focus--notransition:focus {
```
akdetrick: tiniest nitpick ever - would be good to make the variant camelCase to match other classes: `border--focus--noTransition`

src/base-styles/helperClasses/border/focus.css
```diff
@@ -0,0 +1,24 @@
+/*---
+title: Focus Style
+name: focusStyle
+section: Border Helpers
+---
+
+Adds standard focus style to elements
```
akdetrick: might be nice to clarify in docs that the class is selecting the `:focus` state. I can see someone wondering why they don't see the focus outline by just adding the class

src/base-styles/helperClasses/animations/transitions.css
rey-wright: yeah exactly. I think you have the right thought, but we're not in the right phase to really think about multiple transitions... it's one of those things where we'd add it, but if designers or engineers don't use it, then it's wasted. I'd rather someone ask for a different transition at some point, and then we could even deprecate transition--default in a major FDS release, but I'd rather do that than create a taxonomy that either isn't approved by design, or take the time "to" approve it.

rey-wright:
@akdetrick alright, going to go ahead and merge this. If you see anything I did that seems wrong, very ok with you tweaking it while I'm gone. Thanks for the review.

PR #201, opened 2019-12-24
129/components/uip 441 toaster transition rtg

akdetrick:
## Description
Moves `Toaster` transition into a generic set of CSS classes to use with `react-transition-group`

## Checklist
- [x] All unit tests pass


akdetrick:
The only real candidate for using this transition right now is `Toaster`. Just generalizing it so the next time we need a "slide in from the bottom" transition, it will match the toaster

PR #214, opened 2020-01-03
UIP-548: Conditionally render Avatar as a button

rey-wright:
## Description
- We need to sometimes render Avatar as a button or an anchor, in addition to a div
- I modified `baseElement` that I'm using in other components to enable this functionality. I think the approach I used is really slick and allows for a lot of fine grained variability.

## Screenshots
Screen Shot 2020-01-03 at 2 41 53 PM

## Checklist
- [x] All unit tests pass


src/util/baseElement.js
```diff
@@ -4,14 +4,19 @@
  * override where you can specify any element / component as
  * the root
  */
-const baseElement = ({ href, as }) => {
+const baseElement = ({ href, onClick, as }) => {
   let Element;
   if (as) {
     Element = as;
+  } else if (href && onClick) {
+    // This may seem unnecessary but needs to be here to enable the right priority
```
akdetrick: `href` isn't a valid `button` property - should `baseElement` be using an anchor when `href` is passed?

rey-wright:
@akdetrick yeah that's an incredibly good call. I changed the logic, and it ended up making this a lot more simpler (so I was able to kill the comment).

PR #215, opened 2020-01-06
Chip documentation cleanup

rey-wright:
## Description
- removed size="lg" prop from InlineBlockList chip examples (size doesn't exist)
- moved `noop` to general utility folder for reuse (I believe we're using this elsewhere but not going to search for that now)
- reorganized proptype definitions in order of importance (which will reflect in Storybook

## Screenshots
N/A

## Checklist
- [x] All unit tests pass


src/components/util/noop.js
```diff
@@ -0,0 +1,2 @@
+const noop = () => {};
```
akdetrick: `() => {}` is es6 convention for a noop. Not sure we need a utility for this, but I'm not against it either

rey-wright: hmmm so I agree in a sense... but usually when I do noop's, I do them inline. The fact that you broke it out into a variable, was a signal to me that we'd have multiple noop variables in different files, and I'd just break it out into a utility. I guess my follow up question would be... do we need variables for this?

akdetrick: probably not - if I stored a noop in a var, I probably shouldn't have

rey-wright: yeah. You have noops defined in both this story, and the popover component. I'll wipe them away (if that's what you want)

rey-wright: actually... the subsequent one in Popover may have been Nico.

PR #216, opened 2020-01-06
UIP-502: Storybook accessibility

rey-wright:
## Description
Adding Storybook accessibility addon

## Screenshots
Screen Shot 2020-01-06 at 2 29 54 PM

## Checklist
- [x] All unit tests pass


rey-wright:
😄we really need some automated test to verify that IE11 is still working. I wonder how easy it would be to add something like that.

rey-wright:
@akdetrick I don't believe it breaks it. I want to test the deployed version to have the last final word, but I don't believe the accessibility addon itself should break IE11.

rey-wright:
Yep, confirmed this all works in IE11.

PR #219, opened 2020-01-07
130/components/uip 547 checkbox

akdetrick:
## Description
Adds `Checkbox` component.

## Screenshots
Screen Shot 2020-01-07 at 12 14 54 PM
Screen Shot 2020-01-07 at 12 15 16 PM

## Checklist
- [x] All unit tests pass


contributing/naming-conventions.md
```diff
@@ -74,7 +74,8 @@ Props for event handlers should include an `on` prefix:
 +onHover
 ```
 
-Boolean props should have an auxiliary verb prefix:
+Boolean props should have an auxiliary verb prefix _unless_ the boolean prop maps directly
```
akdetrick: @rey-wright How do you feel about this?

contributing/naming-conventions.md
```diff
@@ -74,7 +74,8 @@ Props for event handlers should include an `on` prefix:
...
+to a boolean html attribute:
```
akdetrick: @rey-wright how do you feel about this?

akdetrick:
forgot the knobs...

contributing/naming-conventions.md
rey-wright: @akdetrick yep this is it. We sort of already know we do this (for instance, we don't make `isDisabled` props, especially if the element already has `disabled` natively, as it's yet another thing that the user has to "learn". I sort of like the verb to mostly communicate that it's a boolean, "because" you aren't already familiar with this prop. We don't have that issue with existing boolean props that map onto the underlying html element because the developer already knows those.

rey-wright:
Screen Shot 2020-01-09 at 2 44 20 PM

My contention here is that this is maybe interesting to you or I, but is largely noise to the consumer. All they want is a Checkbox, and they want to know how to use it. Accessibility (that they don't have to interact with) and implementation aren't of much concern to non contributors.

rey-wright:
Screen Shot 2020-01-09 at 2 46 27 PM

Could this be something we'd encapsulate in the jsdoc above that required field? It's right at the top there anyway.

src/components/forms/Checkbox.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+import CheckEmptyIcon from '../../../lib/icons/react/CheckEmptyIcon';
+import CheckFilledIcon from '../../../lib/icons/react/CheckFilledIcon';
+import CheckIndeterminateIcon from '../../../lib/icons/react/CheckIndeterminateIcon';
+
+/**
+ * @param {Object} props react props
+ * @returns {ReactElement}
+ */
+const Checkbox = ({ name, label, value, onChange, indeterminate, checked, disabled }) => {
```
rey-wright: No rest props for this one?

src/components/forms/Checkbox.jsx
```diff
@@ -0,0 +1,93 @@
...
+    Icon = CheckIndeterminateIcon;
+  } else if (isChecked) {
+    Icon = CheckFilledIcon;
+  }
+
+  return (
+    <div className={cx('fdsCheckbox', { 'fdsCheckbox--disabled': disabled })}>
+      <input
+        ref={(el) => {
+          /* eslint-disable no-unused-expressions,no-param-reassign */
```
rey-wright: can we do `// eslint-disable-next-line no-unused-expressions,no-param-reassign` here? Would that work?

src/components/forms/Checkbox.jsx
```diff
@@ -0,0 +1,93 @@
...
+import cx from 'classnames';
+
+import CheckEmptyIcon from '../../../lib/icons/react/CheckEmptyIcon';
+import CheckFilledIcon from '../../../lib/icons/react/CheckFilledIcon';
+import CheckIndeterminateIcon from '../../../lib/icons/react/CheckIndeterminateIcon';
+
+/**
+ * @param {Object} props react props
+ * @returns {ReactElement}
+ */
```
rey-wright: Are you manually typing this, or is this generated? I sort of wonder the degree of helpfulness of this... but if it is helpful, maybe we can make some standards around it?

src/components/forms/Checkbox.jsx
```diff
@@ -0,0 +1,93 @@
...
+        id={name}
+        value={value}
+        className="media--xs"
+        onChange={handleChange}
+        checked={isChecked}
+        disabled={disabled}
+      />
+      <label className="flush--bottom" htmlFor={name}>
+        <div role="checkbox" aria-checked={isChecked}>
+          <Icon size="xs" />
```
rey-wright: Sweet ol' reference pattern.

src/components/forms/Checkbox.jsx
```diff
@@ -0,0 +1,93 @@
...
+      </label>
+    </div>
+  );
+};
+
+Checkbox.defaultProps = {
+  value: undefined,
+  checked: false,
+  indeterminate: false,
+  disabled: false,
```
rey-wright: are the defaultProps here for explicitness, or specifically because undefined doesn't work in some expected way.

src/components/forms/checkbox.stories.mdx
```diff
@@ -0,0 +1,73 @@
...
+      <div>
+        <Checkbox disabled checked name="disabled-checked" label="Disabled checked" />
+      </div>
+    </React.Fragment>
+  </Story>
+</Preview>
+
+## Indeterminate checkbox
+
+HTML checkboxes [natively support an indeterminate state](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Using_checkbox_inputs).
```
rey-wright: Again, I sort of wonder here if this is actually helpful for consumers. This seems like something that would go in the playbook (sans the documentation). I actually have a playbook section in the FDS v7 organizational PR

src/components/style/forms/checkbox.css
```diff
@@ -0,0 +1,40 @@
+/* stylelint-disable selector-max-combinators */
+/* stylelint-disable selector-max-compound-selectors */
```
rey-wright: Lol cheating.... (jk)

src/components/forms/Checkbox.jsx
```diff
@@ -0,0 +1,93 @@
...
+ * @returns {ReactElement}
+ */
+const Checkbox = ({ name, label, value, onChange, indeterminate, checked, disabled }) => {
+  const [isChecked, setIsChecked] = useState(checked);
+
+  const handleChange = () => {
+    if (!disabled) {
+      const newState = !isChecked;
+      setIsChecked(newState);
+      onChange(newState);
```
rey-wright: very very small nit: is there a better variable for `newState`? Like...  `updatedCheckedState` or `isCheckedUpdated` or something more contextual?

src/components/forms/Checkbox.jsx
akdetrick: good point - for consistency with other form components, we should spread on the input

src/components/forms/Checkbox.jsx
akdetrick: It's just something that gets carried over when I start a new component. There's a low chance it will ever be useful, but lots of tools play well with JSDoc so I figure it can't hurt.

As for standards, this is just optional JSDoc.

src/components/forms/Checkbox.jsx
akdetrick: I really love it now

src/components/forms/Checkbox.jsx
akdetrick: I felt it was easier to reason about. It's really only there to be explicit.

src/components/forms/checkbox.stories.mdx
akdetrick: I think you're right - this specific example probably isn't useful to many people (if any).

src/components/style/forms/checkbox.css
akdetrick: I would if it worked. I've resorted to doing this until I have time to further investigate why line-specific comments are ignored by the linter

src/components/forms/checkbox.stories.mdx
akdetrick: A few thoughts on explanatory content in these docs generally...

I think you're right that the majority of users will only care about basic API docs for a component. There will be some users (maybe product, design, or curious engineers) that want to get a better understanding of the intention of a component or how it works.

The thing I love most about docblocks is that you're able to serve varying personas:

> I just need the API docs

look at the props table, skip the rest

> I'm doing something advanced with the component

read more of the examples in the docs (mostly `Preview` stuff)

> Why is this prop required?

read the explanations to get a better understanding

> I learn better by gathering more context

read the extra text for context

With all that said, I don't think this note about native support for indeterminate checkboxes is very useful. I only added that in response to a few misunderstandings about indeterminate states I heard from designers and engineers last month

akdetrick:
Added `Info` component for MDX info boxes for now. We can replace with the official component when available.

Screen Shot 2020-01-10 at 6 55 41 PM

akdetrick:
As for the `source` component, I think it's helpful to see the JSX at a glance.

I'll organize the `Info` util when I get to `Radio`

PR #220, opened 2020-01-08
🔧 Prevents body click listener when popover not click controlled

sfrieson-cbi:
## Description
Fixes an issue with controlled popover still being affected by body clicks.
the branch `popover-bugs` has a story reproducing the bug.

## Checklist
- [x] All unit tests pass


rey-wright:
@sfrieson-cbi what's the bug here?

sfrieson-cbi:
There's a reproduction branch: `popover-bugs`. There's a story in the Popovers folder that explains it.

rey-wright:
@sfrieson-cbi https://github.com/cbinsights/form-design-system/compare/master...popover-bugs

The branch `popover-bugs` is identical to master.

Just to give a small take: for the PR description / commit message, I find it more helpful in general to know the "why", rather than the "what". I can sort of see the "what" was done, in general from the diff, but it takes slightly more effort for uninvolved developers with the fix, on code review, to guess why:

Screen Shot 2020-01-09 at 11 28 49 AM

I'm assuming the body click is responsible for changing the state that shouldn't be changed when the component is uncontrolled, and that state changes X, causing X bug. But I don't really know the bug here, I just know what you did.

To be fair, you didn't assign me to code review 😅and I could probably find out what the bug is if I play with it in uncontrolled mode (or look at this branch) but in general I think a description of "why" would give more context for a quick approval, rather than having to go check out those branches to figure out the why myself.

rey-wright:
Just to explain that a little more:

https://github.com/cbinsights/form-design-system/pull/222

For this PR, I could have said "change position relative to position fixed" but that wouldn't explain the why... only the what. The developer code reviewing this would have no context for why I changed this, unless they were previously involved in the conversation.

sfrieson-cbi:
I'm sorry that the repro branch didn't work properly, but I feel like the description of this pr says the why:

> Fixes an issue with controlled popover still being affected by body clicks.

My issue was that I had a controlled component that was being affected by body clicks.

It seems like your asking for the how, steps for reproduction, which seem more much appropriate for a issue ticket than a PR.

sfrieson-cbi:
The reproduction branch is successfully pushed up.  The description of the bug from the story is below. I hope this expedites the process for you.

___

This version of the popover has `interactionMode="controlled"` and it `isOpen` if the input has content. This works as expected if you click the button in the popover when it shows. If you type into the text area, then click the body to leave it, then click the text area again the popover will not show. Typing more does not get it to show either. It will only show if you clear out the text in the input and start typing again.

```jsx
export const ControlledPopoverWithState = () => {
  const [value, setValue] = useState('');
  return (
    <Popover
      interactionMode="controlled"
      isOpen={value}
      trigger={(
        <input value={value} onChange={(e) => setValue(e.target.value)} />
      )}
    >
      <button onClick={() => setValue('')}>{value}</button>
    </Popover>
  )
};
```


This is a similar attempt at the same functionality but using `interactionMode="click"` and conditionally rendering the popover. You can reproduce the bug in several different ways here. Generally trying different combinations of the steps above will yield unexpected results.

```jsx
export const UncontrolledPopoverWithState = () => {
  const [value, setValue] = useState('');
  return (
    <Popover
      trigger={(
        <input value={value} onChange={(e) => setValue(e.target.value)} />
      )}
    >
      {value
        ? <button onClick={() => setValue('')}>{value}</button>
        : <div />
      }
    </Popover>
  )
};
```

akdetrick:
Because this changes the behavior of the controlled interaction mode, I recommend [bumping the minor version](https://github.com/cbinsights/form-design-system#updating-version-number).

This change means that in controlled mode, the developer is 100% responsible for managing open/close states. That includes clicking away from the popover or hitting escape. That may be a lot of boilerplate to write for every controlled popover, but maybe that's the point of a fully controlled component...

rey-wright:
@sfrieson-cbi appreciate the explanation! Generally we've been following a pattern of elaborating on the "why" in our commit messages or PR title / description. I'll Slack you directly to discuss.

sfrieson-cbi:
The effects are less severe in a later version of FDS.

PR #230, opened 2020-01-15
UIP-568: Badge

rey-wright:
## Description
I spy... a cool badge we can pull into FDS 🙃 

## Screenshots
Screen Shot 2020-01-15 at 10 58 55 AM

## Checklist
- [x] All unit tests pass


src/components/media/Badge.stories.mdx
```diff
@@ -0,0 +1,32 @@
+import { Meta, Story, Preview, Props } from "@storybook/addon-docs/blocks";
+import { withKnobs, text, optionsKnob as options } from '@storybook/addon-knobs';
+import { arrayToOptions } from '../util/storybook';
+
+import Badge, { SIZES } from './Badge';
+
+<Meta title="Interactive|Badge" decorators={[withKnobs]} component={Badge} />
+
```
akdetrick: Good opportunity to communicate design intention here. What kind of thing is this used for, generally. Based on looks alone, this doesn't seem much different than a chip

rey-wright: Yeah it's a good question. I feel like it may be easiest to deploy it, and then once it's there, have design answer to it (and then encode that answer in the documentation), rather than wait for an answer from them on their reasoning for introducing this new UI. I've learned, especially given that I have to catch these things, it's just easier to push it into Storybook first and get the context later.

PR #233, opened 2020-01-17
UIP-572: Button accepts label instead of children now

rey-wright:
## Description
Right now Button accepts arbitrary JSX, so developers could put unapproved things inside Button. Realistically this should only ever be a string. Because of the convention of children generally allowing JSX, I've switched the prop to be a label, and now it must be a string (it's also a required field now).

This is a breaking change, hence why it's being done in `release/7.0`.

## Screenshots
N/A

## Checklist
- [x] All unit tests pass

### DIFF
--- a/src/components/interactive/stackedButton.test.js
+++ b/src/components/interactive/stackedButton.test.js
@@ -1,23 +1,19 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 
-import Button from './Button';
+import StackedButton from './StackedButton';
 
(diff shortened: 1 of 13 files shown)


src/components/interactive/stackedButton.test.js
```diff
@@ -1,23 +1,19 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 
-import Button from './Button';
+import StackedButton from './StackedButton';
```
akdetrick: 😌

PR #245, opened 2020-01-24
Add Styleguide and Intro doc

rey-wright:
Working pretty good for now!
Screen Shot 2020-01-24 at 12 59 36 PM


akdetrick:
would this help?
https://github.com/storybookjs/storybook/pull/6472/files#diff-8129faa75562fa6f6ff738730379d12aR78

rey-wright:
@akdetrick funny enough it DOES, but not really in the way you'd expect! I'll commit a change I'm working on, and then explain later.

rey-wright:
Screen Shot 2020-01-24 at 12 59 36 PM

So this is how I need to do this for now (which is more than acceptable IMO)

akdetrick:
i wonder if it's possible to modify this in the `loadStories` function we define in the config

akdetrick:
update: i just looked at the code and we were thinking the same thing

rey-wright:
@akdetrick so the issue here, is that the staggered importing only works "within a section". The true fix is getting help from the `storySort`, which I already knew about, but concluded it wouldn't help us. The super super lucky part here, is that the `storySort` provided in the snippets floating around issues, sorts alphabetically, and it "just so happens" that FDS section comes alphabetically before the other sections. For that reason, I put the styleguide in that section, and we're also lucky the word "intro: also comes alphabetically... else it would be sorted all wrong. I'm 100% sure you can alter how the sections are organized, but it's very daunting how it works, let alone how the storysort function we're using works, and no one wants to even come close to answering questions about it. So we're super lucky this just happens to work for us alphabetically for now.

PR #248, opened 2020-01-27
Add ability to use absolute paths in FDS

rey-wright:
## Description
- `src` is the root resolution folder. So `../media/Updatable'` becomes `components/media/Updatable`.
- I added an alias for the react icons. so `'../../../lib/icons/react/FeedIcon'` now becomes `icons/FeedIcon`.
- I have not tested if this affects the build, but my guess is that it doesn't. This fully works in Storybook (including hot reloading)
- I didn't go convert all paths. Wanted to make sure this was something we were interested in first.

## Screenshots
Screen Shot 2020-01-27 at 10 52 11 AM

## Checklist
- [x] All unit tests pass


babel.config.js
```diff
@@ -11,5 +11,15 @@ module.exports = {
...
     ['@babel/preset-react'],
   ],
-  plugins: ['@babel/plugin-transform-modules-commonjs'],
+  plugins: [
+    '@babel/plugin-transform-modules-commonjs', 
+    [
+      "module-resolver", {
+        "root": ["./src"],
+        "alias": {
+          "icons": "./lib/icons/react",
```
akdetrick: nice! I think we have two options here...

We can just alias `src` as root and `lib`:

```diff
-import Flex from '../layout/Flex';
-import FDS from '../../../lib/dictionary/js/styleConstants';
-import CheckIcon from '../../../lib/icons/react/CheckIcon';
+import Flex from 'components/layout/Flex';
+import FDS from 'lib/dictionary/js/styleConstants';
+import CheckIcon from 'lib/icons/react/CheckIcon';
```

**OR**

We can drill down the `lib` aliases further (like you've already started with `icons`):

```diff
-import Flex from '../layout/Flex';
-import FDS from '../../../lib/dictionary/js/styleConstants';
-import CheckIcon from '../../../lib/icons/react/CheckIcon';
+import Flex from 'components/layout/Flex';
+import FDS from 'dictionary'; // resolves to `lib/dictionary/styleConstants`
+import CheckIcon from 'icons/CheckIcon'; // resolves to `lib/icons/react/`
```

An alias is already a little bit "magic", so I'm leaning toward going all the way and making import paths as convenient as possible. What do you think?

rey-wright: hmm a lot to unpack here. To be brief, I think your former solution and reasoning makes sense, and can be done with the least amount of work and thought. I think we should go ahead with that for now 👍 thanks for the suggestion.

PR #251, opened 2020-01-29
UIP-606: added white background to outlined version, updated Button text colors

rey-wright:
## Description
Design team wants outlined version to be solid white.

- I updated the interaction states based on the Abstract file
- I bumped up the button color text to primary to solve contrast ratio issues
- I also did this on request for both StackedButton and IconButton

## Screenshots
Screen Shot 2020-01-29 at 6 37 33 PM

## Checklist
- [x] All unit tests pass


src/components/style/interactive/button.css
```diff
@@ -141,6 +153,7 @@
 .inverted .fdsButton--ghost,
 .inverted .fdsButton--outlined {
   color: var(--color-white);
+  background: transparent;
```
akdetrick: So on light colors, the ghost and outline buttons will have white backgrounds, and on dark backgrounds, they will be transparent?

Maybe more of a design question, but are we sure we want a `ghost` button on a `sky` background to be white?

rey-wright: @akdetrick I think I've resolved all the issues here, and Asaf has addressed them in a way that I think makes sense. Essentially, all primary color themes no longer have transparency. You can now consider the outlined button to be a `white` button (similar to blue). It's not necessary used on inverted contexts. The only other true tranparent buttons will be the monochrome, and ghost ones. I've made the necessary modifications, but happy to show my storybook and talk through it with you.

src/components/interactive/button.stories.mdx
```diff
@@ -182,7 +182,7 @@ to take more than a couple of seconds of loading time
 
  ### Inverted
 
-We currently supported inverting of both `ghost` and `outlined` theme
+We currently supported inverting of `ghost` theme
```
akdetrick:
```suggestion
We currently support inverting the `ghost` theme
```

src/components/style/interactive/button.css
```diff
@@ -99,32 +99,44 @@
 
 .fdsButton--outlined {
   border: 1px solid var(--button-outlined-border);
+  background-color: white;
+  color: var(--font-color-dark);
```
akdetrick:
```suggestion
  color: var(--font-color-primary);
```

I don't think `dark` exists any more

src/components/style/interactive/button.css
akdetrick: For clarity, would it make sense to put this `background: transparent` up at line 106 where `ghost` is initially declared, or does it need to be here for the cascade

PR #258, opened 2020-02-13
133/components/uip 197 menubutton

akdetrick:
## Description
Adds `Menu` component using `@reach/menu-button`.

## Screenshots
Screen Shot 2020-02-13 at 3 55 05 PM
Screen Shot 2020-02-13 at 3 55 19 PM
Screen Shot 2020-02-13 at 3 55 42 PM

## Checklist
- [x] All unit tests pass


src/components/popovers/__snapshots__/menu.test.js.snap
```diff
@@ -0,0 +1,141 @@
...
+            "registerDescendant": [Function],
+            "unregisterDescendant": [Function],
+          },
+          "_currentValue2": Object {
+            "descendants": Array [],
+            "registerDescendant": [Function],
+            "unregisterDescendant": [Function],
+          },
+          "_threadCount": 0,
+          "displayName": "MenuDescendantContext",
```
rey-wright: This stuff in the snapshot isn't the greatest.

src/components/popovers/menu.stories.mdx
```diff
@@ -0,0 +1,136 @@
...
+Renders a basic keyboard navigable dropdown menu following the [WAI-ARIA pattern for "menubutton"](https://www.w3.org/TR/wai-aria-practices-1.1/#menubutton).
+
+### When to use Menu
+Only use a `Menu` for dropdowns that contain actions the user can take. If a dropdown includes any other type of content (forms, arbitrary content), use a `Popover` component instead.
+
+### Positioning
+The menu will automatically position above or below the trigger element based on its position in the viewport.
+
+<Story name="Knobs" parameters={{ docs: { disable: true }}}>
+  <div style={{ padding: '2rem' }}>
```
rey-wright: This is a nitpick, but do we want to be using rem here?

src/components/style/modals/menu.css
```diff
@@ -0,0 +1,29 @@
...
+/* use highlight color instead of focus ring */
+[data-reach-menu-item]:focus {
+  outline: 0; /* stylelint-disable-line */
+}
+
+[data-reach-menu-item][data-selected] {
+  background: var(--color-lightGray);
+  color: inherit;
+}
+
```
rey-wright: One note here: all the components in Reach UI seem to spread props, so the className could be provided on the components. I don't have a strong preference here though.

src/components/style/index.css
```diff
@@ -11,6 +11,9 @@
 /* Variables */
 @import "./vars/buttonColors.css";
 
+/* Reach framework styles */
+@import "@reach/menu-button/styles.css";
```
rey-wright: How much of what they define here are we actually using?

src/components/popovers/menu.stories.mdx
```diff
@@ -0,0 +1,136 @@
...
+
+<Props of={Menu} />
+
+## MenuItem Props
+
+<Props of={MenuItem} />
+
+## Trigger element
+
+The Menu trigger can be any JSX, but a `Button` with the `hasCaret` prop enabled is recommended.
```
rey-wright: Do we want to recommend the DropdownButton be used here instead (as well as use that component as the example component throughout this story)?

src/components/popovers/__snapshots__/menu.test.js.snap
akdetrick: Using a `mount` instead of `shallow` so that the `Menu` _and_ `MenuItem` components are included in the snapshot. This does a deep render which includes a lot of the stuff the library includes.

There are really three ways I can think of that this component can break due to code changes:
- className mistakenly removed
- internal nesting structure of `Menu` changes
- the library is updated (potentially with breaking api changes)

In each of these scenarios, the snapshot should break. Doing the deep render helps guarantee that will happen.

src/components/popovers/menu.stories.mdx
akdetrick: no reason, really - just wanted to arbitrarily position the knobs story away from the edge of the canvas so it's easier to see the dynamic positioning.

src/components/style/modals/menu.css
akdetrick: I'm following the styling guide they published, which recommends using these attribute selectors. It's a total gamble, but my bet is that this method of styling will most likely remain stable in future updates of the library. I could be wrong, though

src/components/style/index.css
akdetrick: I considered that, but we want to stay in sync with the library as it gets updated.

Part of the api design for Reach is to include base styles for each component, so Reach should be the point of truth for those base styles instead of us defining a separate one. Our file is for overrides only.

src/components/popovers/menu.stories.mdx
akdetrick: Over the last few weeks, my thinking on how we classify these popovers has shifted. Might be worth a deeper discussion. Basically, when in doubt, I think we should draw the lines using WAI-ARIA guidelines.

### Menubutton pattern
A menubutton is a button that triggers a menu. The menu is a list of actions to be taken. There is no "current selection", because this pattern is not about selecting a value from a list of options.

### Select pattern
A select pattern is a button or input that triggers a list of items that can be selected (they are not actions, just values to be selected). The button or input that triggers the selection list should display some information about the current selection.

The `DropdownButton` accepts arbitrary JSX because it must support showing something about a current selection. It also resembles an input more than a button, which is appropriate for a select-like pattern. It should be baked into some kind of `CustomSelect` component that mimics native `select` interaction. We may want to consider making `DropdownButton` a private component that only a `CustomSelect` would use.

The `Menu` component is following the menubutton pattern, which means the button that triggers the menu does not change based on actions taken within the menu. There is no "current selection" to show. That's why I'm recommending a plain `Button` here.

rey-wright: Makes sense 👍 I'm going to add something to components weekly about this as this may have unforeseen design ramifications.

PR #262, opened 2020-02-19
UIP-272: dictionary value previews

rey-wright:
## Description
Provide previews for Dictionary Values

## Screenshots
Screen Shot 2020-02-19 at 2 45 08 PM
Screen Shot 2020-02-19 at 2 45 13 PM

## Checklist
- [x] All unit tests pass

### DIFF
--- a/src/components/util/Table.jsx
+++ b/src/components/util/Table.jsx
@@ -27,18 +27,31 @@ const mostReadableConfig = (hexName) =>
 export const TableCell = (props) => {
   const [copiedText, copyToClipboard] = useClipboard();
 
-  // Checks to see if a color string is passed, and sets it to a background color to be applied
   const background =
-    typeof props.children === 'string' &&
-    (props.children.startsWith('#') || props.children.startsWith('rgba'))
+    props.cellType &&
+    (props.cellType.startsWith('FONT_COLOR') || props.cellType.startsWith('BORDER_COLOR'))
       ? props.children
       : undefined;
+
+  const color = mostReadableConfig(background || '#FFF');
+
+  const fontFamily =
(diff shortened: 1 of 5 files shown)


src/components/util/Table.jsx
```diff
@@ -27,18 +27,31 @@ const mostReadableConfig = (hexName) =>
...
+
   return (
     <td
       style={{
+        fontFamily,
+        fontWeight,
+        fontSize,
         background,
-        color: mostReadableConfig(background || '#FFF'),
+        color,
```
rey-wright: I thought about doing a conditional and some other things but the way I have it seemed the most straightforward. I'm up for suggestions on how I could simplify the logic here.

akdetrick: What's wrong with `mostReadable` to set the color?

I think this is fine either way though; we can always iterate on it if need be.

rey-wright: oh I 'm mostly referring to all the conditional logic: 

```  
const fontFamily =
    props.cellType && props.cellType.startsWith('FONT_FAMILY') && props.children;

  const fontSize =
    props.cellType && props.cellType.startsWith('FONT_SIZE') && props.children;

  const fontWeight =
    props.cellType && props.cellType.startsWith('FONT_WEIGHT') && props.children
```

Essentially for each variable, I'm doing duplication in that I'm checking for the cellType, and then checking that it starts with a certain value, and then assigning props.children to that value, so that it can be applied via shorthand. I could get rid of the duplicated check by maybe doing a function that returns an object with these keys, and then spread that, and then just make sure they're populated when cellType is there... but that's too in the weeds maybe.

PR #263, opened 2020-02-20
[v8] Checkbox

rey-wright:
## Description
Refactored Checkbox

## Screenshots
N/A

## Checklist
N/A


akdetrick:
Before we dig into any details, what is motivating this change from an uncontrolled component to a controlled component?

rey-wright:
@akdetrick I'm not really sure 😄 

If we can change it back to uncontrolled and certain functionality is preserved, that's a desirable outcome. It was just easier for me to copy Reach UI's code examples just to mostly talk about functionality / api.

src/components/forms/Checkbox.jsx
```diff
@@ -1,78 +1,69 @@
...
-  const id = uuidv4();
+import { CustomCheckboxInput, CustomCheckboxContainer } from '@reach/checkbox';
 
-  const IconUnchecked = CheckEmptyIcon;
-  const IconChecked = indeterminate ? CheckIndeterminateIcon : CheckFilledIcon;
+const Checkbox = ({ label, defaultChecked, showLabel = true, ...props }) => {
+  const [checkedState, setChecked] = useState(props.checked || defaultChecked || false);
+  const checked = props.checked != null ? props.checked : checkedState;
+
+  // We should maybe fire onChange in "addition" to setChecked, based on whether we
```
akdetrick: not sure if this helps resolve the question, but the main difference to the user is passing `checked` (controlled) vs `defaultChecked` (uncontrolled).

src/components/forms/Checkbox.jsx
```diff
@@ -1,78 +1,69 @@
...
+  // interpret checkbox as being "controlled" or not
+  const onChange = props.onChange ? props.onChange : (e) => setChecked(e.target.checked);
```
akdetrick: an `onChange` prop is required when this is acting as a fully controlled component.

In other words, if the user passes `checked`, they become fully responsible for updating the value of that prop from outside the component.

If `defaultChecked` is passed, then the component can manage checked state itself.

src/components/forms/Checkbox.jsx
```diff
@@ -1,78 +1,69 @@
...
-      />
-      <label className="flush--bottom" htmlFor={id}>
-        <span className="fdsCheckable-icon--checked">
-          <IconChecked size="xs" />
-        </span>
-        <span className="fdsCheckable-icon--unchecked">
-          <IconUnchecked size="xs" />
+    <Element className="display--inlineBlock">
+      <CustomCheckboxContainer
+        checked={props.checked != null ? props.checked : checked}
```
akdetrick: This is where things get tricky - does the Reach component mandate that this input be fully controlled? If not, we should pass through `checked` or `defaultChecked` props instead of always setting `checked` from here

akdetrick:
I'm on board with these changes; there are just a few things to consider with controlled vs uncontrolled.

There are some benefits to allow the component to do either based on what the end user passes in.

### Controlled
Useful if a parent component is using internal state or redux state to explicitly set checkbox `checked` state

```
<Checkbox
	label="Agree"
	onChange={() => { setIsAgreeChecked(!isAgreeChecked) })
	checked={isAgreeChecked}
/>
```

### Uncontrolled
For situations where we need only to listen to change events and respond to them.
```
<Checkbox
	label="Agree"
	onChange={(e) => { /* do something based on e.target.value */})
	/* user can optionally set `defaultChecked` here to set a default state */
/>
```

rey-wright:
@akdetrick thanks for the feedback. Some of those comments may be old. 

I think the key for me, is that all of the Checkbox story permutations I added work. Should we talk through this some more just to make sure it's what we want? Or are we fine with it.

What I'm probably going to want to do here, once you approve, is to go into cbi-site and try to replace most if not all the checkboxes with a beta version of this, and throw that on staging... I want to make 100% sure this Checkbox meets our needs. If that works, I'll keep those changes on cbi-site somewhere around in a commit, we can get this into v8, and then once v8 drops proper, I can apply those changes to cbi-site master.

src/components/forms/checkbox.stories.mdx
```diff
@@ -93,6 +93,37 @@ renders. Read the `checked` attribute of the event `currentTarget`.
...
+          <Checkbox label="Controlled Unchecked" checked={value} onChange={() => setValue(!value)}/>
+        )}
+      />
+      <br />
+      <Checkbox label="Uncontrolled" />
+      <br />
+      <Checkbox label="Uncontrolled Default Checked" defaultChecked />
+      <br />
+      <Checkbox id="blah" />
+      <label className="display--inlineBlock" htmlFor="blah">Custom label (htmlFor)</label>
```
akdetrick, line 126: should probably remove the custom label example

src/components/forms/Checkbox.jsx
```diff
@@ -1,72 +1,69 @@
...
-        className="media--xs"
-        disabled={disabled}
-        ref={inputRef}
-        {...rest}
-      />
-      <label className="flush--bottom" htmlFor={id}>
-        <span className="fdsCheckable-icon--checked">
-          <IconChecked size="xs" />
+    <Element className="display--inlineBlock">
+      <CustomCheckboxContainer
```
akdetrick, line 33: looks like this still needs to support the `disabled` class/state

akdetrick:
Also needs a `CHANGELOG` entry under v8

PR #265, opened 2020-02-25
UIP-628: add dialog features (and stories covering them)

rey-wright:
## Description
Added a bunch of requested dialog features that were discussed with engineers, design (asaf) and TPM's. Once this is featured reviewed, I can get the code ready for code review.

* Added `subsection` (inserts arbitrary content underneath header)
* Added `alwaysShowBorder` (to always show footer border)
* Added `height` (so developers can add a custom height, similar to what we do for width now)

Open questions:

* What do we want to expose to Prompt? I'm thinking that we shouldn't expose any of these new props to it, but it's up for debate.

## Screenshots
N/A

## Checklist
- [x] All unit tests pass


src/components/modals/Dialog.jsx
```diff
@@ -71,7 +71,14 @@ const Dialog = (props) => {
...
                     onKeyDown={handleKeyDown}
-                    style={{ maxWidth: `${props.width}px` }}
+                    style={{
+                      maxWidth: `${props.width}${
+                        typeof props.width === 'number' ? 'px' : ''
+                      } `,
+                      maxHeight: `${props.height}${
+                        typeof props.height === 'number' ? 'px' : ''
+                      }`,
+                    }}
```
rey-wright: I'm thinking we shouldn't accept numbers in the future, to allow developers to supply their own units. That being said, I added this code to detect numbers and add pixels, and if it's not a number, to not add anything, so that we wouldn't unnecessarily introduce a breaking change here.

src/components/modals/Dialog.jsx
```diff
@@ -104,7 +116,9 @@ const Dialog = (props) => {
                     {props.footerContent && (
                       <div className="dialog-footer">
                         <Section
-                          border={isOverflowing ? 'top' : undefined}
+                          border={
+                            props.alwaysShowBorder || isOverflowing ? 'top' : undefined
+                          }
```
rey-wright: It "might" be worth adding some conditional checks that are computing `isOverflowing` and not run them if `props.alwaysShowBorder` is set (ideally this is a rare prop, but could be a nice add).

src/components/modals/Dialog.jsx
akdetrick: This is nice.
If we begin accept strings, we should also update the proptypes and their jsdoc comments

CHANGELOG.md
```diff
@@ -27,6 +27,11 @@ This is an example of a brief overview of the _Major_ or _Minor_ version changes
 
 ---
 
+## [7.8] Updated Dialog
```
akdetrick: another conflict to resolve (sorry)

src/components/modals/Dialog.jsx
akdetrick: would definitely be nice to avoid the overflow checks when possible

src/components/modals/Dialog.jsx
rey-wright: I updated the prop type to say that it accepts strings, with the assumption that numbers are still supported, but it's sort of a soft deprecation to not list it, and then we'll rip it out in the next major.

src/components/modals/Dialog.jsx
```diff
@@ -152,7 +162,16 @@ Dialog.propTypes = {
...
 
   /** Custom modal width */
-  width: PropTypes.number,
+  width: PropTypes.string,
+
+  /** Custom modal height */
+  height: PropTypes.string,
+
+  /**
+   * Border currently renders only if there is scrollable content. Set this boolean
```
akdetrick:
```suggestion
   * Border between content area and footer currently renders only if there is scrollable content. Set this boolean
```

rey-wright: Gratzie

PR #268, opened 2020-02-27
UIP-625: added TextInput component

rey-wright:
## Description
Added a TextInput component

## Screenshots
Screen Shot 2020-02-25 at 12 17 17 PM

## Checklist
- [x] All unit tests pass


rey-wright:
@akdetrick in reference to our 1:1 today, I'm happy to remove the stories for IconInput and DecoratedInput if that feels like a good move to make once you look through this PR.

src/components/forms/TextInput.jsx
```diff
@@ -0,0 +1,115 @@
...
+  type,
+  IconLeft,
+  IconRight,
+  after,
+  before,
+  ...props
+}) => {
+  // Refer to the following for an explanation on
+  // the throttling / debouncing approach:
+  // https://www.peterbe.com/plog/how-to-throttle-and-debounce-an-autocomplete-input-in-react
```
rey-wright: I highly highly recommend reading this article, just to make sure we're aligned on the approach. There may be different approaches / articles we can look into as well.

There's actually another thing in the article that I might steal... throttling if the length of the characters is less than 5, AND if the ending character ever ends in a space. Article explains why you'd want to do that.

CHANGELOG.md
```diff
@@ -27,6 +27,10 @@ This is an example of a brief overview of the _Major_ or _Minor_ version changes
 
 ---
 
+## [7.8] Add TextInput
```
akdetrick: you'll need to bump this again (merge from master)

src/components/forms/TextInput.jsx
```diff
@@ -0,0 +1,115 @@
...
+  type: 'text',
+  onChange: () => {},
+  throttleOnChange: () => {},
+};
+
+TextInput.propTypes = {
+  errorText: PropTypes.string,
+
+  /** Label used for `label` element */
+  label: PropTypes.string,
```
akdetrick: How do you feel about making this required and having a `showLabel` prop the way we do with current checkbox?

If a developer wants to hide the label, we use the `label` string on an `aria-label` attribute of the input.

src/components/forms/TextInput.jsx
```diff
@@ -0,0 +1,115 @@
...
+   */
+  IconLeft: PropTypes.func,
+
+  /**
+   * FDS Icon _reference_
+   * `IconRight={StarFilledIcon}`
+   */
+  IconRight: PropTypes.func,
+
+  /** Standard input type */
```
akdetrick: minor suggestion

```suggestion
  /** Input element `type` attribute */
```

src/components/style/forms/decoratedInput.css
```diff
@@ -29,7 +29,8 @@
 
   /* use line box to center text and match input height */
   padding: 0 var(--space-half);
-  line-height: var(--space-double);
+  display: inline-flex;
```
akdetrick: 🙏

src/components/forms/textInput.stories.mdx
```diff
@@ -0,0 +1,177 @@
...
+    before="before"
+    after="after"
+    IconLeft={CaretDownIcon}
+    IconRight={CaretDownIcon}
+  />
+</Story>
+
+### Misc
+
+Can be used with InputGroup (why don't these stretch full width?)
```
akdetrick: was this an unaddressed TODO?

src/components/forms/TextInput.jsx
```diff
@@ -0,0 +1,115 @@
+import React, { useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { throttle, debounce } from 'throttle-debounce';
+import IconInput from 'components/forms/IconInput';
+import DecoratedInput from 'components/forms/DecoratedInput';
+
+export const throttleValue = 500;
+
+const TextInput = ({
```
akdetrick: Needs unit tests

akdetrick: Should we remove the stories for `IconInput` and `DecoratedInput` in this branch?

src/components/forms/TextInput.jsx
rey-wright: So yeah I guess this becomes a bit tricky. I can see pros and cons of both sides of this.

If we force the developer to pass a label regardless, then what happens to the visual label that they need to show users? Do they still use their own label (so that there are 2 labels)? Or do they just add a span with some text, and the "true" label is on the `aria-label`? I sort of see the benefit of making label `required` so that devs don't abuse not passing a label (which will 100% happen), but then again I sort of question the developer having to apply "2" labels, the one they pass to the component, and then the one they specify themselves, and then the whole thing where they click on their own label... would that focus the input. Lot of... variables here 🙂

src/components/forms/textInput.stories.mdx
rey-wright: This is possibly more of a question for you I think? IconInput, DecoratedInput, etc, all stretch to be 100% width, except InputGroup. Sort of wondering if it's expected that this is different, or if it should be the same.

I'm glad I wrote this because I completely forgot about it.

src/components/forms/TextInput.jsx
rey-wright: I'm down, as long as you agree / see this component replacing `IconInput` / `DecoratedInput`.

rey-wright: I'll take this question as an implicit thumbs up and go ahead and remove it for now 🙂

src/components/forms/TextInput.jsx
```diff
@@ -0,0 +1,115 @@
...
+  label,
+  errorText,
+  throttleOnChange,
+  required,
```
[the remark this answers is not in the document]
rey-wright: Interesting... 

So what I've been doing personally, is if the boolean is spread directly onto the component as expected, even if it causes additional side effects, I keep it that name. So like... a perfect example is the disabled attribute of button. Technically, anchors don't have a disabled prop, but I figured it made more sense to accommodate for the `button` element that does.

In this change, if input itself uses the required prop as it normally does, I'd at least "historically" choose to keep calling it required, so that the developer knows that the prop carries the same intention as what the normal input would. That being "said", I can see the alternative angle of using the `is` prefix the moment it's being used in any capacity beyond its normal usage.

Essentially, I don't feel our usage guidelines totally encapsulates this specific use case.

All that being said, that's just what I've done historically, and honestly either way would make sense to me. What do you think makes the most sense in this situation?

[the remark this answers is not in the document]
rey-wright: The other side thing to note: funny that you mention it, but I technically did "not" add the required field to the input yet. Are we comfortable with the UI that makes the input border completely red if it's a required field? Because that's what happens to our current input.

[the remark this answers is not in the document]
rey-wright: Screen Shot 2020-03-02 at 1 15 24 PM

Screen Shot 2020-03-02 at 1 16 00 PM

So if required is applied to input, this is how it would look.

src/components/forms/TextInput.jsx
rey-wright: Hmmm so I can do snapshot tests as an easy fix. I could "also" test the throttling functions but I'm hesitant to test that, unless we're confident that we don't want to change the throttling configuration.

src/components/forms/TextInput.jsx
```diff
@@ -0,0 +1,116 @@
...
+import { throttle, debounce } from 'throttle-debounce';
+import IconInput from 'components/forms/IconInput';
+import DecoratedInput from 'components/forms/DecoratedInput';
+
+export const throttleValue = 500;
+
+const TextInput = ({
+  label,
+  errorText,
+  throttleOnChange,
```
rey-wright: @akdetrick how do you feel about `onThrottleChange`?

src/components/forms/TextInput.jsx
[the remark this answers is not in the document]
akdetrick: I suspect that this is the `:invalid` state (required but empty). Does it only happen on blur? If so, it might be okay...

src/components/forms/TextInput.jsx
akdetrick: The pattern would be...

```
label: PropTypes.string.isRequired,
showLabel: PropTypes.bool, // default `true` - if set to `false`, apply `label` to `aria-label`
```

src/components/forms/textInput.stories.mdx
akdetrick: the outer border for an input group is the root element of the `InputGroup` component. This element is set to `min-content` because the last `input` or `select` element in the group does not grow to fill remaining space.

src/components/forms/TextInput.jsx
rey-wright: right but I don't get the surrounding things that happen as a result of this. We should maybe chat about it, it's slightly confusing.

src/components/forms/TextInput.jsx
akdetrick: `onThrottledChange` makes the most sense to me (past tense)

src/components/forms/TextInput.jsx
akdetrick: Snapshots to cover different prop combinations would work. For throttling, maybe spy on `onThrottledChange` - Jest timers may be useful here. If it's too much of a pain, snapshots might be enough.

akdetrick: Yep, I'd vote on removing those from docs

src/components/forms/TextInput.jsx
akdetrick: `onThrottledChange` (past tense Throttled) makes the most sense to me

src/components/forms/TextInput.jsx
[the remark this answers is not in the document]
rey-wright: @akdetrick unfortunately I believe this happens regardless of blur.

src/components/forms/TextInput.jsx
rey-wright: I'm now testing that onThrottldeChange is now firing. I'm opting to test the function I'm using directly to test whether the debounce vs throttle variant of the function fires.

src/components/forms/textInput.stories.mdx
```diff
@@ -0,0 +1,224 @@
...
+    />
+  </Story>
+</Preview>
+
+
+### Controlled Input
+
+<Story name="Controlled Input">
+  <TextInput
+    value="Controlled"
```
akdetrick: our example should also show an `onChange` prop for controlled input

src/components/forms/textInput.stories.mdx
```diff
@@ -0,0 +1,224 @@
...
+    placeholder="placeholder"
+    before="before"
+    after="after"
+    IconLeft={CaretDownIcon}
+    IconRight={CaretDownIcon}
+  />
+</Story>
+
+### Form management library
+You should be able to use this input in any form management library you'd like:
```
akdetrick: I'd like to declare our preference for react-hook-form here

```suggestion
You may use this input in any form management library you'd like. We recommend `react-hook-form`, which is used in the example below:
```

CHANGELOG.md
```diff
@@ -27,6 +27,10 @@ This is an example of a brief overview of the _Major_ or _Minor_ version changes
 
 ---
 
+## [7.10] Add TextInput
```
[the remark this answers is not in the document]
rey-wright: no worries 👍 

In general I usually wait on solving conflicts like these until after approval, as I'll just need to solve them again if not.

src/components/forms/textInput.stories.mdx
rey-wright: thanks for the suggested change 👍

rey-wright:
@akdetrick one note here: I added support for "orientation" if you spin up the server and take a look at the knobs. I "may" need some of your flex expertise to get it working properly, but this can always be addressed later via a patch / minor update.

src/components/forms/TextInput.jsx
```diff
@@ -0,0 +1,143 @@
...
+  /** Custom onChange event that throttles / debounces. */
+  onThrottledChange: PropTypes.func,
+
+  /** String to place to the left of the input */
+  before: PropTypes.string,
+
+  /** String to place to the right of the input */
+  after: PropTypes.string,
+
+  orientation: PropTypes.oneOf(['vertical', 'horizontal']),
```
akdetrick: Only thing left to address here - I suspect the `horizontal` orientation may more commonly need the label on the left of the input.

This also opens up a potential API change if we need to give users control over left/right/top

rey-wright: @akdetrick added!

rey-wright:
@akdetrick Oh I forgot to ask my 2nd question: I added some interesting flex stuff to this PR to accommodate for the orientation changes. Wondering if you could take a look at the flex CSS I added to see if it makes sense (and if there's any chance adding those changes could break anything). This is mostly around changing CSS for things like fdsDecoratedInput and the like.

rey-wright:
I'll resolve conflicts on this when it gets approved.

rey-wright:
@akdetrick 

- `orientation` is now `labelPosotion`
- `left` and `top` are the only valid values now. `right` has been removed.

PR #269, opened 2020-02-28
UIP-634: storybook configuration migration, addressing sorting, updating to 6.0 alpha

rey-wright:
## Description
Couple of important points and play by play:

- Storybook has a good system in that it gives deprecation warnings in minor releases (and has an upgrade guide here: https://github.com/storybookjs/storybook/blob/master/MIGRATION.md)
- You're able to migrate away from the deprecation warnings without upgrading to the major, which allows for a more smooth upgrade to the major when it's released.
- Storybook has been giving warnings about separators on the dev server. I adjusted all of those. The separator deprecation "requires" you to give a section to all stories.
- The new Component section, because it alphabetically comes before "FDS", required me to implement sorting on sections, which I finally figured out.
- While I was here, I ended up apply some of the general sections in the Storybook taxonomy ticket (note: I did not completely tackle reorganizing of components, as it requires a breaking change, just added some of the section sorting).
- I haven't figure out explicit page sorting, but I can figure that out down the road.
- There was another change in terms of configuration files that will be made, so I adjusted those files and configuration as well.
- This all worked without me having to update the major
- There was one small bug regarding the FDS logo which requires upgrading to 6.0 alpha, and "only" for that reason. The alpha looks pretty good, so decided to upgrade.
- With the upgrade to configuration here, the storybook instance "starts" slower in development. However, it reloads faster, and switches stories faster.

## Screenshots
Screen Shot 2020-02-27 at 7 14 16 PM

## Checklist
- [x] All unit tests pass


.storybook/preview.js
```diff
@@ -0,0 +1,35 @@
...
+    showRoots: true,
+    storySort: (a, b) => {
+
+      if (a[1].kind === b[1].kind){
+        return 0;
+      }
+
+      const sectionSortOrder = ['intro', 'ui', 'design', 'icons', 'components'];
+
+      const sortOrder = (() => {
```
akdetrick: this is a good use case for `reduce`

```
const sortOrder = ['intro', 'ui', 'design', 'icons', 'components']
	.reduce((ordering, curr, i) => {
		ordering[curr] = i;
		return ordering;
	}, {});
```

PR #275, opened 2020-03-05
130/components/uip 383 datepicker

akdetrick:
## Description
Adds `DateInput` component.

## Screenshots
Screen Shot 2020-03-05 at 5 27 42 PM
Screen Shot 2020-03-05 at 5 27 49 PM
Screen Shot 2020-03-05 at 5 28 53 PM

## Checklist
- [ ] All unit tests pass


src/components/forms/DateInput.jsx
```diff
@@ -0,0 +1,212 @@
...
+ */
+export const isValidUserDate = (inputValue) => VALID_DATE_REGEX.test(inputValue);
+
+/* eslint-disable react/prop-types */
+/**
+ * Private component for DateInput that renders Month/Year select inputs.
+ * @param {Object} props react props
+ * @returns {ReactElement}
+ */
+const YearAndMonthSelector = ({ date, localeUtils, onChange, startYear, endYear }) => {
```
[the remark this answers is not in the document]
rey-wright: No strong feelings on that particular point. I'd generally keep them in the same file if it was a small file. The only thing here is that it seems this file is quite long though, which may merit moving "something" to another file (no idea what that something is though).

src/components/forms/DateInput.jsx
```diff
@@ -0,0 +1,212 @@
...
+
+    if (isValidUserDate(value)) {
+      const parsedDate = moment(value, DATE_FORMAT);
+      setPickerMonth(parsedDate.toDate()); // update the picker month/year in picker as user types
+      handleDaySelect(parsedDate.toDate());
+    }
+  };
+
+  return (
+    <React.Fragment>
```
rey-wright: the nitpickiest of nitpicks: Fragment isn't needed here? Also, I've been trying more and more to use `<></>` instead of `<React.Fragment></React.Fragment>`

src/components/forms/DateInput.jsx
```diff
@@ -0,0 +1,212 @@
...
+  onDateChange: PropTypes.func,
+
+  /** Number of years into the past to show in the year dropdown */
+  pastYears: PropTypes.number,
+
+  /** Number of years into the future to show in the year dropdown */
+  futureYears: PropTypes.number,
+
+  /** Default date selection */
+  defaultDate: PropTypes.instanceOf(Date),
```
rey-wright: 👍

src/components/forms/DateInput.jsx
```diff
@@ -0,0 +1,212 @@
...
+  // Expand range to include selected year if out of range
+  if (selectedYear && selectedYear < startYear) {
+    startYear = selectedYear;
+  }
+  if (selectedYear && selectedYear > endYear) {
+    endYear = selectedYear + 1;
+  }
+
+  return { startYear, endYear };
+};
```
rey-wright: This is personal preference maybe, but functions like this that have this many arguments, it's usually not immediately clear what the function is doing in use, vs if you use an object.

e.g.

```
getYearRange(2020, 3, 5) // I have no idea what 3 and 5 are, without knowing that the function signature is
getYearRange({ year: 2020, pastYears: 3, futureYears: 5}) // this is more clear on glance
getYearRange(2020, { pastYears: 3, futureYears: 5}) // this is also somewhat clear as well.
``

File: src/components/forms/dateInput.test.js [hunk 5]
```diff
@@ -0,0 +1,98 @@
...
+      yearSelect = wrapper.find('select').at(1);
+    });
+
+    afterEach(() => {
+      changeFn = null;
+      wrapper = null;
+      input = null;
+      monthSelect = null;
+      yearSelect = null;
+    });
```
rey-wright: 💪

src/components/style/forms/dateInput.css
```diff
@@ -0,0 +1,67 @@
...
+  right: var(--space-half);
+}
+
+.fdsDateInput .DayPicker-wrapper {
+  padding: 0;
+}
+
+.fdsDateInput .DayPicker-Caption {
+  padding-top: var(--space-half);
+  margin: -8px;
```
rey-wright: Negative margins 😉Welcome to the dark side

src/components/forms/DateInput.jsx
```diff
@@ -0,0 +1,212 @@
...
+import moment from 'moment';
+
+import DayPicker from 'react-day-picker';
+
+import ActionsArrowLeftIcon from 'lib/icons/react/ActionsArrowLeftIcon';
+import ActionsArrowRightIcon from 'lib/icons/react/ActionsArrowRightIcon';
+import Popover from '../popovers/Popover';
+import Flex from '../layout/Flex';
+import FlexItem from '../layout/FlexItem';
+import IconButton from '../interactive/IconButton';
```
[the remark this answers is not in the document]
rey-wright: To be clear 😄 definitely not following this rule in the codebase right now, as I opted to not go and update all the paths when we introduced absolute paths, so I'm planning on doing that later. We could leave this like this for now and I can change it here though.

src/components/forms/DateInput.jsx
```diff
@@ -0,0 +1,212 @@
...
+ * @param {Object} props react props
+ * @returns {ReactElement}
+ */
+const YearAndMonthSelector = ({ date, localeUtils, onChange, startYear, endYear }) => {
+  const months = localeUtils.getMonths();
+
+  const years = Array.from({ length: endYear - startYear }, (year, i) => i + startYear);
+
+  const handleChange = (e) => {
+    const { year, month } = e.target.form;
```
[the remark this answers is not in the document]
akdetrick: good catch - will clean that up

src/components/style/forms/dateInput.css
```diff
@@ -0,0 +1,67 @@
...
+}
+
+.fdsDateInput select {
+  font-size: var(--font-size-s);
+  height: var(--fdsDateInput-control-height);
+}
+
+.fdsDateInput .fds-icon:last-child {
+  margin-left: 4px;
+}
```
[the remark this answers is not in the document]
akdetrick: left this in by mistake - removing

src/components/forms/DateInput.jsx
akdetrick: Yeah, I agree with this generally - 4 positional arguments is pretty much the mental limit when I break out args into object properties. If this were exposed in docs and intended for usage outside of this component, I would definitely favor an object here.

In this case, this function covers logic that I factored out of `DateInput` to make it a pure function. Above all else, I wanted this function to be "pure". I preferred positional arguments to do this because...

- only the last param is optional
- explicitly declaring the arguments keeps the surface area of the function interface as small as possible (objects can contain _anything_)
- accepting an object means you could have unrelated parameters being passed in
- positional args discourage doing "unpure" things like mutating the object being passed in
- easier to jsDoc params this way

src/components/style/forms/dateInput.css
akdetrick: 🙈

src/components/forms/DateInput.jsx
akdetrick: *tl;dr* - positional arguments are "safer" if functional purity is the goal, object arguments are nicer for public interfaces. `getYearRange` is not intended to be a public interface

akdetrick:
**note to self:** render "outside days" as hint color

akdetrick:
most recent commit enables "outside days" - useful when someone picks a day in the first or last week of a month and then navigates forward or back a month

Screen Shot 2020-03-11 at 4 15 05 PM

PR #276, opened 2020-03-05
UIP-549: FDS styles

rey-wright:
## Description
Added Style Utilities (I didn't name it Styling Utilities as the word wraps and it doesn't look great when it wraps).

Github collapses the diffs of the mdx files so you'll want to open those to see what's going on.

## Screenshots
Screen Shot 2020-03-05 at 5 49 37 PM

## Checklist
- [x] All unit tests pass
- [ ] ~**[Version bumped](https://github.com/cbinsights/form-design-system#updating-version-number) if appropriate**~


rey-wright:
Fixed all lint issues.

rey-wright:
Screen Shot 2020-03-05 at 6 51 03 PM

Note: there are a bunch of pages that "technically" you could argue that they would be more streamlined if the examples moved into the table. This an example of a place where that could happen. It could also happen for Media and a lot of other places. I've sort of been looking at this for too long and I think it's fine either way now, but happy to accommodate any requests to put any of these examples inside of the table.

rey-wright:
There's similar questions here. I could take these colors and insert them into the actual table, that way users can see context for the color. We could also put them in the table, AND leave them out of the table. We could also put a link to dictionary to see a more pretty representation of our colors if they want to see the colors broken out. Lot of options here.

Screen Shot 2020-03-06 at 10 19 07 AM

rey-wright:
Screen Shot 2020-03-05 at 5 49 37 PM

There's another idea here as well. Because this list is so long, and obscures the component list, we "could" put all of these stories in single folder. People can still search for specific things they need, but the list wouldn't be expanded by default like it is now. Another idea.

src/components/intro.docs.stories.mdx
```diff
@@ -10,6 +10,7 @@ https://github.com/cbinsights/form-design-system
 
 - [📖 Dictionary](https://cbinsights.github.io/form-design-system/fds-dictionary)
 - [🎨 Base Styles](https://cbinsights.github.io/form-design-system/fds-styles)
+- [⚛️ React Components](#components-forms)
```
rey-wright: This link doesn't actually work right now. I need to fix this (it's simple, I just don't want to make any more commits on this for now)

akdetrick: separate PR works for me

akdetrick:
locally everything looks good - I don't have any strong arguments for or against rolling up styles stories into a folder. I'm not as concerned with things being "below the fold" because people scroll pretty naturally and search is still a nice way to find what you're looking for

.storybook/preview-head.html
```diff
@@ -64,10 +64,21 @@
...
+
+  .doctable tr:last-child td:last-child, .doctable tr:last-child th:last-child {
+    border-bottom-right-radius: 4px;
+  }
+
+  .doctable tr:first-child td:last-child, .doctable tr:first-child th:last-child {
+    border-top-right-radius: 4px;
+  }
+
   .doctable td {
```
akdetrick: very minor, maybe for a follow up.

What do you think about putting a light border between table rows?

rey-wright: should be there in Chrome. Are you looking at this in Safari? This is really the exact same table as the normal prop table... the rows may not be there in Safari.

docs/styles/position.stories.mdx
```diff
@@ -0,0 +1,91 @@
...
+      ]} 
+    />
+    <div style={containerStyles}>
+      <Item label="center--center"><div className="alignChild--center--center" style={positionStyles}><div style={boxStyles}/></div></Item>
+      <Item label="center--top"><div className="alignChild--center--top" style={positionStyles}><div style={boxStyles}/></div></Item>
+      <Item label="center--bottom"><div className="alignChild--center--bottom" style={positionStyles}><div style={boxStyles}/></div></Item>
+      <Item label="left--center"><div className="alignChild--left--center" style={positionStyles}><div style={boxStyles}/></div></Item>
+      <Item label="left--top"><div className="alignChild--left--top" style={positionStyles}><div style={boxStyles}/></div></Item>
+      <Item label="left--bottom"><div className="alignChild--left--bottom" style={positionStyles}><div style={boxStyles}/></div></Item>
+      <Item label="right--center"><div className="alignChild--right--center" style={positionStyles}><div style={boxStyles}/></div></Item>
```
akdetrick: I can see arguments both for and against what I'm about to suggest...

Each one of these examples repeats most of the same code. Do we want to stick with the declarative, repetitive pattern, or have each MDX file declare some custom example wrapper?

For example, in this MDX file we could do...

```
const AlignChildExample = ({ x, y }) =>
	<Item label={`${x}--${y}`}><div className={`alignChild--${x}--${y}`} style={{...generalStyles, etc}}><div style={{height: 30, etc}} /></div></Item>;
```

Then in the example, we can just do
```
<AlignChildExample x="left" y="top" />
<AlignChildExample x="left" y="center" />
etc
```

docs/styles/layout.stories.mdx
```diff
@@ -0,0 +1,81 @@
+import { Meta, Story } from "@storybook/addon-docs/blocks";
```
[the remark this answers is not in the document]
rey-wright: Forgot that you brought this up before. This change will be there in the next commit that I push up.

src/components/util/storybook.js
```diff
@@ -109,3 +109,40 @@ export const storyBackgrounds = {
...
+};
+
+export const labelStyle = {
+  fontSize: 11,
+  marginBottom: 10,
+  fontFamily: FDS.FONT_FAMILY_MONO,
+  color: '#f09',
+};
+
+export const Item = (props) => (
```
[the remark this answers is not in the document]
rey-wright: Totally.

docs/styles/styledocUtils.js
```diff
@@ -0,0 +1,86 @@
...
+    <div style={typeStyle}>type--data</div>
+    <div className="type--data margin--bottom--half">Sphinx of black quartz, judge my vow</div>
+    <div style={typeStyle}>type--caption</div>
+    <div className="type--caption margin--bottom--half">Sphinx of black quartz, judge my vow</div>
+    <div><span style={typeStyle}>type--link </span> (anchor)</div>
+    <a className="type--link margin--bottom--half display--block" href="#">Go to Homepage</a>
+  </>
+);
+
+export const InvertedTypesetting = () => (
```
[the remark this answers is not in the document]
rey-wright: I did just notice that `type--big`, `type--date`, and `type--caption` all support inversion. So I can add those. But I think the issue is that the `type--head` classes do not support inversion. If I merely just add a prop to change the inversion and added the inverted class, then the `type--head` classes are going to look weird on the navy background, and I feel it won't be clear to the user.

Maybe I can make this component you're talking about only contain the things that can be inverted / non inverted, and then separate the head classes from them?

This is complicated enough that we should maybe chat about this via Slack whenever you have the bandwidth.

docs/styles/styledocUtils.js
```diff
@@ -0,0 +1,86 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import FDS from 'lib/dictionary/js/styleConstants';
+
+export const generalStyles = {
```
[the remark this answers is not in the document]
rey-wright: ehhh so this is really difficult, and was probably one of the more difficult parts of creating these examples.... essentially, which styles are shared, which styles serve as good defaults, etc. If you notice, initially I made some of these layout components in`styledocUtils` that I use "sometimes"... but quickly notice there were instances where I needed to use the raw styles.

It's tough because I sort of don't want to make component abstractions until I'm "sure" that they're a worthy abstraction. I borderline wanted to go and rip out the components as even the styles aren't that consistent and I end up overriding them to the point where I question their value and general applicability.

But yeah it's rough because I could go in here and create an abstraction that makes more sense, but it's like... 

1. How much time does it makes sense to dedicate to that, given the nature of the work.
2. How many of these abstractions will even stick around? So I'm increasingly pondering the utility of putting the examples INSIDE the tables, so that users don't have to mentally draw a line between the the class, and the example below it that repeats the label. If that eventually happens, it will nullify all of these styles.
3. Even if I create the right abstraction, I'm not totally confident, in this scenario, that it will hold up to scrutiny if we add further examples.

Given all that, I sort of just left it in a mediocre, pretty good, not great but not bad state given the ephemeral state of the current docs. Happy to change it though if you're interested in it.

docs/styles/position.stories.mdx
rey-wright: Some of this is answered by this comment: https://github.com/cbinsights/form-design-system/pull/276#discussion_r390375109

Essentially to summarize:

* This pattern I'm probably only using for like 2 stories, so I feel like the abstraction isn't great, which is why you see me using styles a lot.
* If some of the examples move to inside the table (which I feel I could make a great argument for, and I could complete some of that when I get fatigued from whatever bigger project I'm working), it nullifies both these styles and components. I figure better to put it out there, figure out what users like, get this initiative done, and then slowly layer onto it later. 

But yeah I sort of don't see the benefit in creating an abstraction around this now. To do that, I'd probably would need to make a component that supports the 6 or so different variations of examples I'm using, that again may go away at some point.

docs/styles/styledocUtils.js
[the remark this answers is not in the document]
rey-wright: We resolved online. We want to invert all the classes, and there's a fix incoming to enable inversion support on `type--head` classes. I've resolved this accordingly, and should look good when the patch lands.

docs/styles/styledocUtils.js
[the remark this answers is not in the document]
akdetrick: Sounds like not worth the effort for this round

docs/styles/position.stories.mdx
akdetrick: To be clear, I'm not advocating for a general use example component, just defining a context-specific component when helpful (in the same file it's used).

This is optional though, I don't want to hold this up with over-optimizing things

akdetrick:
**RE: bgColor docs** - They _could_ be in the table, but I don't see anything wrong with the example as it is.

akdetrick:
Inverted heading fixes just merged in, in case you want to squeeze in the inverted `Typography` component prop before merging this PR

docs/styles/position.stories.mdx
rey-wright: oic. Yeah I could do that. I guess the only question there, is if we'd want to make a custom component like that for other places. The value of accepting the direction as a prop IMO is only useful given that the label is still there, and I'm sort of thinking these examples "may" fit better in the table. I'll punt on it for now if it's ok with you, but definitely want to revisit this later on.

PR #277, opened 2020-03-06
UIP-505: importpath

rey-wright:
## Description
Added ImportPath to all documentation components.

This looks better on stories that have the code snippet (Only `Button` right now). When we eventually come back and add this, these stories will look a whole lot better.

## Screenshots
Screen Shot 2020-03-06 at 3 19 01 PM

## Checklist
- [x] All unit tests pass

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -293,4 +293,23 @@
     background-clip: content-box;
   }
 
+  .importPath {
+    color: #333;
+    font-size: 14px;
+    cursor: pointer;
+    font-family: "Operator Mono","Fira Code Retina","Fira Code","FiraCode-Retina","Andale Mono","Lucida Console",Consolas,Monaco,monospace;
+  }
+
+  .importPath--highlight {
+    color: #006699;
+  }
+
+  .importPath:hover, .importPath:hover svg {
+    color: darkorange;
+  }
+
+  .importPath:hover .importPath--highlight {
+    color: orange;
+  }
+
 </style>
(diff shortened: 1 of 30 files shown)


.storybook/preview-head.html
```diff
@@ -293,4 +293,23 @@
...
+  }
+
+  .importPath:hover, .importPath:hover svg {
+    color: darkorange;
+  }
+
+  .importPath:hover .importPath--highlight {
+    color: orange;
+  }
+
```
rey-wright: To note: I'm going to start specifically not using FDS values in here. I want all the stuff in Preview to be more generalized, in the off chance that we want to share some of this stuff back to the Storybook community.

akdetrick: makes sense - I was _just_ about to suggest `var(--font-family-mono)`, but your approach makes sense 👍

src/components/util/storybook.js
```diff
@@ -109,3 +111,43 @@ export const storyBackgrounds = {
   check:
     'url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAcAAAAHCAYAAADEUlfTAAAAHElEQVQYV2M8fPiwLwMOwAiStLW13YxNftBJAgAx2BqeI9XcBAAAAABJRU5ErkJggg==)',
 };
+
+export const ImportPath = ({ component, section }) => {
+  const [, toggleHover] = useToggleHover();
```
akdetrick: minor question - if you only need the second index of a hook return value, is it better to just do this?

```
const toggleHover = useToggleHover()[1];
```

rey-wright: Yeah it's a good question. I feel like a developer could get more easily confused at that, given the overwhelming pattern of hooks to de-structure the array? I can update though... don't personally feel strongly about either way (trade off IMO)

PR #284, opened 2020-03-11
Add base theme to Storybook

kchen-cbi:
## Description
Without a base theme set, Storybook checks the system to determine what the fallback should be. Since our components are not designed to support dark mode, we should override this behavior.

## Checklist


rey-wright:
You'll need to merge in latest master and rebuild again to solve the conflicts here.

PR #285, opened 2020-03-11
UIP-589: set default color for form field

rey-wright:
## Description
Setting default color on inputs. Chrome gives these inputs black, so that's the color that I chose.

## Screenshots
N/A

## Checklist
- [x] All unit tests pass


src/base-styles/base/formTags.css
```diff
@@ -53,6 +53,7 @@ input[type="file"] {
   background: var(--color-white);
   min-width: var(--space-double);
   min-height: var(--space-double);
+  color: var(--color-black);
```
akdetrick: should be charcoal, which is also "primary" text color:

```suggestion
  color: var(--font-color-primary);
```

PR #286, opened 2020-03-11
add snippets to all component stories except layout

rey-wright:
## Description
Add expanded snippets to all stories.

The only section I didn't add is for layouts... there's something about how I'd implement that, that just doesn't feel right to me. All of the component snippets so far are about "simplicity", but because the nature of layout components is to show complexity, I sort of don't know if it makes sense to just adapt the examples from the knobs into a story. There's just too much to unpack there and may be better suited to be addressed at a later time.

## Screenshots
Screen Shot 2020-03-11 at 3 05 33 PM

## Checklist
- [x] All unit tests pass

### DIFF
--- a/src/components/interactive/IconButton.jsx
+++ b/src/components/interactive/IconButton.jsx
@@ -81,9 +81,9 @@ IconButton.propTypes = {
   /** Controls whether the button is disabled or not. */
   disabled: PropTypes.bool,
   /** Used to render a FDS Icon (should only be used for FDS Icons) */
-  Icon: PropTypes.func,
+  Icon: PropTypes.func.isRequired,
   /** Accessibility label */
-  label: PropTypes.string,
+  label: PropTypes.string.isRequired,
 };
 
 export default IconButton;
--- a/src/components/popovers/popover.stories.mdx
+++ b/src/components/popovers/popover.stories.mdx
@@ -28,7 +28,7 @@ import Button from '../interactive/Button';
 
 Base popover component. Bring your own trigger and popover content.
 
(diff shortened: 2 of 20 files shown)


src/components/interactive/IconButton.jsx
```diff
@@ -81,9 +81,9 @@ IconButton.propTypes = {
   /** Controls whether the button is disabled or not. */
   disabled: PropTypes.bool,
   /** Used to render a FDS Icon (should only be used for FDS Icons) */
-  Icon: PropTypes.func,
+  Icon: PropTypes.func.isRequired,
```
akdetrick: 🙏

src/components/popovers/popover.stories.mdx
```diff
@@ -54,6 +54,25 @@ Base popover component. Bring your own trigger and popover content.
...
+<Preview isExpanded>
+  <Popover
+    position='right'
+    alignment='start'
+    distance={4}
+    interactionMode='click'
+    transitionName="GrowFast"
+    trigger={<div className="margin--bottom"><Button label="Click Me" /></div>}
+  >
+    <div
```
akdetrick: we can simplify this example by removing the styled div here

rey-wright: My only concern is that all the simplified stories work as a live example. If we put nothing in here, Popover will seem to not work. If we put just text in here, it'll get pushed all the way to the right side of the screen.

rey-wright:
Happy to make that follow up change but I'd want to discuss that with you first. Going to go ahead and merge this for now though.

PR #289, opened 2020-03-13
UIP-657: remove defaultprops

rey-wright:
## Description

React plans to deprecate defaultProps in favor of having developers use default parameters. Let's migrate now so we don't have to worry about it in the future.

Storybook didn't used to spell out the default parameters in the docs, but now that they do, it makes sense to migrate.

## Screenshots
n/a

## Checklist
- [x] All unit tests pass
- [ ] **~[Version bumped](https://github.com/cbinsights/form-design-system#updating-version-number) if appropriate~**


rey-wright:
One really bad part here: shallow rendering no longer captures the default props inside the component. But that's just a side effect of shallow rendering and snapshots.... mounting the component is going to be more implementation free from refactors and further changes like this anyway.

PR #290, opened 2020-03-13
UIP-658: only display a role of img when the Avatar isn’t either an anchor or a button

rey-wright:
## Description
Right now this flags an accessibility error in Storybook when it renders as a Button or an anchor while still displaying the img role.

I'm not going to bump the minor here... I believe this is just a patch and developers don't really need to know about this at this point.

## Screenshots
Screen Shot 2020-03-13 at 2 02 51 PM

## Checklist
- [x] All unit tests pass


kchen-cbi:
Not sure if it's within the scope of this ticket, but just curious why we're using `role="img"` instead of a native `img` element here. I assume there's a good reason for it, so in that case I would just say we should also add an `aria-label`.

kchen-cbi:
Actually, we need an `aria-label` regardless, since the button/link usage will need it. Also, would probably want it exposed as a prop (maybe with default value of something like `"User Avatar"`), since it should be customized depending on the intended `onClick` behavior of the button.

rey-wright:
@kchen-cbi thanks for the feedback!  It's in scope.

So the reason we're using role="img" is because Avatar may not necessarily render an image.

Screen Shot 2020-03-17 at 11 28 30 AM

So by default, we only require the person to pass a name, which is assigned to the title attribute, and used to generate initials and a circle with a color. We consider this to sort of be an image, just that it's convenient to create with css. I "could" just conditionally render an actual `` but I figure the img role covers us in both cases. Thoughts?

The `aria-label` piece is a good idea... I'll add that right now.

rey-wright:
@kchen-cbi 2 notes:

1. We currently spread props, so if the developer passed an aria-label, it would be correctly applied.
2. People can render these Avatars for the purpose of companies, and I may not want to necessarily couple the business logic. Do you think, given that, that it's still a good idea to display just "Avatar" by default? Or to leave it up to the user to give something more accurate. My fear, give your presentation, is that just specifying "Avatar" is sort of abusing the purpose of what `aria-label` should be, but I'm definitely comfortable deferring to you on this.

kchen-cbi:
> So by default, we only require the person to pass a name, which is assigned to the title attribute, and used to generate initials and a circle with a color. We consider this to sort of be an image, just that it's convenient to create with css. I "could" just conditionally render an actual `` but I figure the img role covers us in both cases. Thoughts?

Ahh, I see. Thanks, makes sense! Not much interactivity lost by not using native `img` so not a big deal anyway.

> 2. People can render these Avatars for the purpose of companies, and I may not want to necessarily couple the business logic. Do you think, given that, that it's still a good idea to display just "Avatar" by default? Or to leave it up to the user to give something more accurate. My fear, give your presentation, is that just specifying "Avatar" is sort of abusing the purpose of what `aria-label` should be, but I'm definitely comfortable deferring to you on this.

I feel like "Avatar" is generic enough that it still works for companies, no? And I think having a default is preferable. Better to have some description of the purpose of the image even if it's missing the specificity of say the user/company name.

rey-wright:
@kchen-cbi makes sense. I've put those changes into effect, let me know what you think.

kchen-cbi:
Looks good to me, thanks!

PR #300, opened 2020-03-19
UIP-669: run lint suite on prepush

rey-wright:
## Description
We want to run the lint suite just in case we were no-verifying commits up until the push.

## Screenshots
N/A

## Checklist
N/A


akdetrick:
nice

PR #301, opened 2020-03-25
UIP-656: Adding popover ref props

rey-wright:
## Description
Added popover ref props: https://cbinsights.atlassian.net/browse/UIP-656

## Screenshots
N/A

## Checklist
- [x] All unit tests pass


src/components/popovers/Popover.jsx
```diff
@@ -24,6 +24,38 @@ export const getPopperPlacement = (position, alignment) => {
   return `${position}${variation}`;
 };
 
+let overflowStyle = null;
+
+const useDisableScroll = (disableScrollRef, isDisabled) => {
+  useEffect(() => {
+    if (disableScrollRef) {
+      const domNode = disableScrollRef.current;
```
akdetrick: this is annoying, but there's more than one way a ref can be defined now (old way, useRef) - fall back directly to the ref reference if `current` is undefined is a way to support both

```suggestion
      const domNode = disableScrollRef.current || disableScrollRef;
```

src/components/popovers/Popover.jsx
```diff
@@ -24,6 +24,38 @@ export const getPopperPlacement = (position, alignment) => {
...
+      if (isDisabled) {
+        overflowStyle = domNode.style.overflow;
+        domNode.style.overflow = 'hidden';
+      } else if (overflowStyle !== null) {
```
akdetrick: might be able to simplify if `overflowStyle` is set outside the conditional (unless you're trying to avoid reflows by reading the property only if `isDisabled` is true)

src/components/popovers/popover.stories.mdx
```diff
@@ -107,3 +108,74 @@ Base popover component. Bring your own trigger and popover content.
...
+          </div>
+        </div>
+      )
+    }}
+  </Story>
+</Preview>
+
+## Close on Scroll
+
+If you'd like the ability to close the popover once the user scrolls, you can utilize `closeOnScrollRef`:
```
akdetrick: does this help or make it harder to understand?

```suggestion
If you'd like the ability to close the popover once the user scrolls some element, you can pass a ref of that element to the `closeOnScrollRef` prop.
```

src/components/popovers/popover.stories.mdx
```diff
@@ -107,3 +108,74 @@ Base popover component. Bring your own trigger and popover content.
     </Popover>
   </Story>
 </Preview>
+
+## Disable Scrolling
+
+If you'd like to disable scrolling when a popover is open on any given element, you can utilize `disableScrollRef`:
```
akdetrick:
```suggestion
If you'd like to disable scrolling of some element when a popover is open you can pass a ref of the scrolling element to the `disableScrollRef` prop.
```

src/components/popovers/Popover.jsx
rey-wright: yeah. There's a specific reason I did this but my mind is shot right now late in the day. I can provide more context when my brain is working tomorrow.

src/components/popovers/Popover.jsx
```diff
@@ -24,6 +24,38 @@ export const getPopperPlacement = (position, alignment) => {
...
+        domNode.style.overflow = overflowStyle;
+      }
+      // We'll want to modify this to accommodate 2 use cases:
+      // 1. The user is using classes to apply scrolling... in this case
+      //    we'd want to clear out all styles when un-disabling.
+      // 2. Overflow might not be just scroll. We should just save whatever
+      //    was there inside a closure here, and then re-assign it back.
```
rey-wright, line 43: note to self: delete these comments

src/components/popovers/Popover.jsx
rey-wright: Yeah so now that I look at this, I can't think of a way that I would want to simplify this, as it might make it more complicated. I think the move would be to demonstrate how this works with a simplified example, and then show why the elseif isneeded here. Given that you mentioned that this is not a blocker, I'm going to move forward with merging, and then glad to chat about it tomorrow so that it makes sense (or if there's a superior way to write or document this).

PR #304, opened 2020-03-26
UIP-644: remove bgColorOptions

rey-wright:
## Description
This should stop the bgColor knob from from appearing on every single story: https://cbinsights.atlassian.net/browse/UIP-644

## Screenshots
N/A

## Checklist
N/A (mdx changes don't require us to do anything else)


rey-wright:
The main note here (which may not be obvious): new storybook does "not like" knobs to be defined outside of components. I solved this by inlining the knob in question here, but I never "removed" the older knob here.

PR #311, opened 2020-04-03
UIP-715: background documentation

rey-wright:
## Description
- Improved background documentation by moving examples inside the table
- This required a long overdue refactor of overcomplicated logic that had to be done to make this possible, which also has the side effect of making copying logic more easy to be used in the future, as well as making working with, and maintaining those table components much easier.

## Screenshots
Screen Shot 2020-04-03 at 4 44 49 PM

## Checklist
ran yarn build:full


akdetrick:
nice!

PR #327, opened 2020-04-16
UIP-738: react popper upgrade

rey-wright:
## Description
Tried to upgrade react-popper but seems broken.

- Breaking changes: https://github.com/popperjs/react-popper/pull/333
- They also "completely rewrote the library" apparently in a minor update, and said file a bug if you spot anything: https://github.com/popperjs/react-popper/releases/tag/v2.2.0 so maybe the move is to try a version earlier than 2.2.0 if figuring out why popover doesn't work here becomes super hard.
- They apparently also now have a popover hook, which we might want to utilize (take a look at previous link for more info)

## Screenshots
Screen Shot 2020-04-16 at 9 08 30 AM

## Checklist
N/A

### DIFF
--- a/src/components/popovers/Popover.jsx
+++ b/src/components/popovers/Popover.jsx
@@ -162,7 +162,7 @@ const Popover = ({
       name: 'offset',
       enabled: true,
       options: {
-        offset: `0,${distance}`,
+        offset: [0, distance],
       },
     },
     {
--- a/src/components/popovers/Popover.jsx
+++ b/src/components/popovers/Popover.jsx
@@ -157,26 +157,27 @@ const Popover = ({
   const clonedTrigger = React.cloneElement(trigger, triggerProps);
 
   // https://popper.js.org/popper-documentation.html#modifiers
-  const popperModifiers = {
-    offset: {
+  const popperModifiers = [
+    {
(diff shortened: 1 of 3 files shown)


src/components/popovers/Popover.jsx
```diff
@@ -157,26 +157,27 @@ const Popover = ({
...
   // https://popper.js.org/popper-documentation.html#modifiers
-  const popperModifiers = {
-    offset: {
+  const popperModifiers = [
+    {
+      name: 'offset',
       enabled: true,
-      offset: `0,${distance}`,
+      options: {
+        offset: `0,${distance}`,
```
FezVrasta, line 165: This should be a tuple

rey-wright: Thanks @FezVrasta, that seems to fix most of the breakage!

rey-wright:
It seems we might be good here, however... it is worth of note that popovers that don't have a lot of space to display, still act weirdly, albeit, in a new "weird" way based on this update. If you look at the popover knobs in the docs view, you'll see what I mean.

rey-wright:
I'm closing this to talk about a different approach we can take in our next 1:1

PR #342, opened 2020-04-29
137/UIP-598 explicit radio props

badakim:
## Description
- Pass props explicitly in the Radio component.
- Add props to the PropTypes.
- Make the input radio readonly when no onChange is passed to resolve the prop type warning.

## Checklist
- [x] All unit tests pass


src/components/forms/Radio.jsx
```diff
@@ -11,7 +11,19 @@ import RadioFilledIcon from 'lib/icons/react/RadioFilledIcon';
...
+      showLabel = true,
+      disabled = false,
+      name,
+      label,
+      checked,
+      defaultChecked,
+      value,
+      onChange,
+    },
+    ref
```
rey-wright, line 25: You'll still want to allow passing rest props here just in case the developer wants to add any thing arbitrary like data attributes, event listeners, whatever (though unlike the current example, could you make sure that `className` comes after the rest props? I want to disable anyone from passing `className` through rest props).

rey-wright:
@badakim this isn't a blocker for the PR, but something I just thought about: 

Now that I think about it, I'm not 100% sure that putting readonly on the radio buttons is correct from an accessibility perspective? That being said, I think I'm in favor of this to get rid of the warning, and then we could revisit it down the line.

badakim:
> @badakim this isn't a blocker for the PR, but something I just thought about:
> 
> Now that I think about it, I'm not 100% sure that putting readonly on the radio buttons is correct from an accessibility perspective? That being said, I think I'm in favor of this to get rid of the warning, and then we could revisit it down the line.

@rey-wright To not have to worry about `readonly`, we could limit the usage to passing `defaultChecked` instead of `checked`, and handle events at the parent level.

PR #347, opened 2020-05-05
add ADR

akdetrick:
ADR for new hierarchy and import structure. **Also updates ADR template** to make ADRs a bit more practical.

We can review at the next OOF meeting before merging.

**Rendered Markdown:** https://github.com/cbinsights/form-design-system/blob/139/adr/package-file-structure/adr/file_organization.md


adr/file_organization.md
```diff
@@ -0,0 +1,80 @@
...
+### `components` file structure
+
+Categorization directories will be removed.
+Each component will have its own directory under `components/`.
+Tests, stories, private components, and helpers will be colocated in component
+directories.
+
+```
+components/
+├── Foo/
```
sfrieson-cbi, line 58: I know this is a little late, but it might be nice in the future to add a test directory and files. An example of how we generally do it in cbi-site

```
├── Foo/
│   ├── __tests__/
│   │   └─ Foo.spec.js
│   └─ Foo.jsx

Comment (rey-wright) [reply adr/file_organization.md:58]: Not too late. Do you think this might be overkill though? Wondering the benefits of providing a folder specifically for tests (especially if it comes pass that we're only writing one test file for our FDS components)

Comment (rey-wright) [reply adr/file_organization.md:58]: Definitely a benefit of alignment though.

Comment (sfrieson-cbi) [reply adr/file_organization.md:58]: In the past I was generally against it, but I'm starting to come around. I think it generally makes things cleaner.
1) It makes it visually easier to find test files when they're all in one dir.
2) It makes it visually easier to find your component file when it's not collated with tests:
```
component-a.jsx
component-a.test.js
component-b.jsx
component-b.test.js
component-c.jsx
component-c.test.js
```
3) It's the Jest suggestion.

These aren't huge reasons. More opinion than anything. I've found easier since working at CBI than my previous job where we organized in the fashion of the example above.

akdetrick:
Add "getting started guide" to ADR

akdetrick:
Consider building main component directly in `index.js` for component dir

rey-wright:
Knowing that we call css `styles.css`, and that we have an `index.js` file, it sort of draws attention to the fact that we don't necessarily need names on the files inside the folder, as the folder serves as the name space. so index.test.js, for instance.

akdetrick:
@rey-wright @reywright  Made some updates to reflect changes in v9 - any reason not to merge this in its current state?

reywright:
@akdetrick So the ADR in it's current form is slightly incorrect as we use the index.jsx pattern inside the folder.

Further than that, I feel like if our index.jsx file is generic, and our css file is generic, then the rest of the files in this folder "should" be generic.

Further than that, I wanted to look into Steven's strategy and think about that more. I think I just got distracted but I want to make sure whatever we come up with, we've thought through any things used on the cbi-site that we potentially would want to bring over.

This probably warrants some verbal discussion.

PR #350, opened 2020-05-05
add naming convention for component reference props:

akdetrick:
It's a bit too late to write a good ADR on the decision background because I've lost a lot of my memory of this discussion.

Moving forward, we can take notes in the ADR format as we're debating the decision.

For this decision, I'll just add a naming convention note to contributing guidelines

### DIFF
--- a/contributing/naming-conventions.md
+++ b/contributing/naming-conventions.md
@@ -115,6 +115,13 @@ Always name rest props `rest`:
 +const Component = ({ foo, bar, ...rest})
 ```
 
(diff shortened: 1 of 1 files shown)


rey-wright:
Yeah I agree. Maybe as we decide one of these things, I can be writing an ADR for it. I definitely feel like you're going to get the language / premise better than I, but I can at least do one, and that's what the code review is for.

akdetrick:
Yep - better as a note taking template than a "write a thing after the fact" sort of thing.

PR #352, opened 2020-05-11
add tooltip delay

akdetrick:
## Description
Adds delay to tooltip.

## Checklist
- [x] All unit tests pass


akdetrick:
Trimmed 300ms off of the delay. Good catch on cancelling; that should be happening now.

CHANGELOG.md
```diff
@@ -27,6 +27,11 @@ This is an example of a brief overview of the _Major_ or _Minor_ version changes
 
 ---
 
+## [8.3] Add delay to `Tooltip`
+
+- Adds support for `delay` prop in base `Popover` component
+- Adds a `1000`ms hover delay to `Tooltip`
```
rey-wright, line 33: Just a gut check: if we end up updating the specific number for the delay here, that may be a patch, so just wondering if the number value should be here vs just specifying we now specify a default hover delay. That being said, that's a very very small nit.

PR #358, opened 2020-05-26
Clarify controlled popover docs

sfrieson-cbi:
## Description
As this is the second time that I have had bugs due to this functionality ([here's an allusion to the first time](https://github.com/cbinsights/form-design-system/pull/220)), I've added a bit to the documentation about the fact that the popover can still close by itself in controlled mode and fleshed out the controlled example to show what is necessary (synchronizing state in the `onClose`) to work with it.

If a user does not synchronize their state they will not otherwise know that the popover has closed, which means they will not be able to open it again if using an open button or they will have to click twice if using a toggle button.

## Checklist
- [x] All unit tests pass


reywright:
Also when I say fixed, I mean fixed by UI Platform team, not in this PR.

sfrieson-cbi:
I agree, but [a comment on the previous pr](https://github.com/cbinsights/form-design-system/pull/220#issuecomment-572702000) that @akdetrick made, brings up the valid concern that this puts the onus on the user to recreate the outside click/escape functionality. This is nice to be given out of the box, so if that is taken away, but I was ready to implement it myself for this component because I feel like that's what it means to be controlled.

My main desire for my use-case was that it could act more like a system menu where I could have a click _inside_ the popover close the popover. Currently controlling the popover is the only way to signal this to the popover component. A simple `shouldCloseOnInsideClick` boolean would not do either because there could be areas that are disabled or not interactive that could be clicked by the end-user. The only API to support this that I've thought of so far is to allow a function that returns a boolean to be passed as `shouldCloseOnInsideClick` or maybe an array of refs to act as close-worthy targets.

Any thoughts?  I'll gladly make a PR to fix the bug in anyway that seems appropriate.

reywright:
@sfrieson-cbi hmmm maybe we should chat. I think the PR could go in now though and any sort of bug fix or feature addition we can add onto later, I think the context is helpful while we don't have this fixed. I added some time tomorrow for us to talk about it. You're free to not join if you're busy but yeah would be great to discuss with you if you're available.

sfrieson-cbi:
We chatted in person. We decided on making bug fixes and changes to the popover API that will make this documentation and example moot in its current state.

PR #363, opened 2020-05-27
UIP-762: jscodeshift

reywright:
## Checklist
- [ ] All unit tests pass


transforms/transform.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
```
akdetrick, line 1: (optional) This could go in `scripts/bash`.

PR #367, opened 2020-06-03
add popoverProps to DateInput

akdetrick:
## Description
Adds `popoverProps` object prop to `DateInput`

## Screenshots
Screen Shot 2020-06-03 at 10 48 17 AM

## Checklist
- [ ] All unit tests pass

### DIFF
--- a/src/components/forms/DateInput.jsx
+++ b/src/components/forms/DateInput.jsx
@@ -170,6 +170,7 @@ const DateInput = ({
   futureYears = 1,
   pastYears = 40,
   dateFormat = 'MDY',
+  popoverProps = {},
   defaultDate,
   onDateChange,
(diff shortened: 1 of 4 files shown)


src/components/forms/DateInput.jsx
```diff
@@ -240,6 +244,7 @@ const DateInput = ({
           IconRight={DatePickerIcon}
         />
       }
+      {...popoverProps}
```
reywright, line 247: I can see both sides of why we would / wouldn't do this, but we sure we want to use rest props here?

akdetrick: `Popover` itself does not spread props on any DOM elements, so I figured this was safe to do here.

PR #369, opened 2020-06-03
forward ref in Popover

akdetrick:
## Description
`Popover` now forwards its `ref` prop to the content element.

**see comment below** - not much code actually changed here.

## Checklist
- [x] All unit tests pass


src/components/popovers/Popover.jsx
```diff
@@ -69,189 +69,196 @@ const useCloseOnScroll = (closeOnScrollRef, isActive, closeCallback) => {
...
+      isOpen,
+      transitionName,
+      disableScrollRef,
+      closeOnScrollRef,
+    },
+    forwardedRef
+  ) => {
+    const [isActive, setIsActive] = useState(false);
+    const refTriggerWrap = useRef(null);
+    const refContent = forwardedRef || useRef(null);
```
akdetrick, line 94: this is the main thing that changed - if a `ref` is passed, we use the `forwardRef` instead of creating a new ref to use internally.

reywright: I think it's against hooks rules to define them conditionally. Could the conditional be defined in a different place?

Screen Shot 2020-06-03 at 3 34 50 PM

akdetrick: i think i "fixed" this by using the `createRef` api for this ref instead

akdetrick:
@arianamestel-cbi @rey-wright Feel free to merge this without me if it passes code review (i'm out until Monday)

PR #371, opened 2020-06-09
add input change callback to date input

akdetrick:
## Description
Adds callback for input change event in `DateInput`

## Checklist
- [x] All unit tests pass


src/components/forms/DateInput.jsx
```diff
@@ -275,11 +277,15 @@ const DateInput = ({
 
 DateInput.propTypes = {
   /**
-   * Change callback for date selection.
+   * Change callback for date selection. This is called when a valid
+   * date has been selected in the pickr or entered into the input.
```
reywright, line 281: nit: `picker`

src/components/forms/DateInput.jsx
```diff
@@ -222,6 +223,7 @@ const DateInput = ({
 
   const handleInputChange = ({ target: { value } }) => {
     setInputValue(value);
+    onInputChange();
```
reywright: should we provide either the value, or the event? e.g. `onInputChange(value)`.

If we would pick one of them, I'd choose the event just to be consistent with the normal `onChange` but not too opinionated here.

PR #378, opened 2020-06-12
[v9] UIP-781: inlineblocklist separator list

reywright:
## Description
Rename InlineBlockList to SeparatorList.

- No change in functionality
- That being said, we want to make it clear that this is primarily for separators
- We will be introducing a component called GapList that specializes in spacing out elements.

## Screenshots
N/A

## Checklist
- [ ] All unit tests pass


CHANGELOG.md
```diff
@@ -57,6 +57,11 @@ This is an example of a brief overview of the _Major_ or _Minor_ version changes
 - `lib/icons/raw/` will no longer exist in packages published v9 or later. For raw SVG
   icons, use `lib/icons/svg/`.
 
+#### Renamed InlineBlockList component
+- `InlineBlockList` component is now `SeparatorList` as we want the naming to have more
```
akdetrick, line 61: (minor) we can explain that difference briefly

```
- `InlineBlockList` is now `SeparatorList`. This component is primarily used for configuring separator characters between items in an inline-block list. An upcoming component `GapList`, will render an inline-block list with configurable spacing between items.
```

akdetrick: i'm not sure about this honestly; your call

reywright: I feel like explaining the difference of a component that is yet to be released doesn't make sense. That being said, I think the fix here is just to release that component in v9, and I'm almost done with it.

reywright: @akdetrick As previously discussed, I think moving a lot of things to v9 helps simplify things. GapList is now set to merge into v9, so I think adding this explanation now makes more sense. I took your explanation and modified it slightly, let me know how it works for you now.

reywright: I'll merge this now, but we can easily make adjustments before v9 is merged.

PR #380, opened 2020-06-15
[v9] UIP-782: GapList component

reywright:
## Description
Adding GapList component to allow users to easily control spacing through a number of elements

Note: the "spacing" PR has been merged into this one, and I'll probably merge this one PR and close that other PR.

## Screenshots
Screen Shot 2020-06-15 at 11 02 04 AM

## Checklist
- [x] All unit tests pass


src/components/GapList/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+    case 3:
+      return 16;
+    case 4:
+      return 24;
+    case 5:
+      return 32;
+    default:
+      return 16;
+  }
+};
```
reywright, line 20: So.... yeah 😅 

- This is for applying the negative margin solution
- I can't use classes as we don't have negative margin variants (and we may only need them in FDS)
- There's really no way I know of for "dynamically" constructing a variable, outside of using `eval`.
- This is sort of one of the bad parts about using consistent jumps. If we had the jumps at 1, 2, 4, 6, and 7, then this function would just be multiplying.
- Let me know if you have any thoughts on this. We can easily make this a utility and use it where we need to, but yeah. Maybe the way here is to use css variables possibly....

akdetrick:
Holding until spacing PR merges

akdetrick:
moved to PR #388

PR #382, opened 2020-06-16
fix ref prop types in dateInput

akdetrick:
fix prop types


src/components/forms/DateInput.jsx
```diff
@@ -322,16 +322,22 @@ DateInput.propTypes = {
...
+    PropTypes.func,
+    PropTypes.shape({ current: PropTypes.any }),
+  ]),
 
   /**
    * Object accepting any valid prop from `Popover` except for `trigger` and `ref`.
    * To apply a `ref` to the popover element, use the `popoverRef` prop.
    */
-  popoverProps: PropTypes.object,
+  popoverProps: PropTypes.func,
```
akdetrick, line 340: nope

PR #384, opened 2020-06-17
[v9] UIP-833: add 2 new props (isError and isLabelBold) to TextInput

reywright:
## Description
- Adding `isError` if users don't pass errorText but still want to get the input into the error ui state
- Adding `isLabelBold` if users want to bold their labels

## Screenshots
Screen Shot 2020-06-17 at 11 05 48 AM

## Checklist
- [ ] All unit tests pass


src/components/TextInput/index.jsx
```diff
@@ -19,8 +19,10 @@ const TextInput = React.forwardRef(
   (
     {
       label,
+      isLabelBold,
```
reywright, line 22: Note: the ticket itself calls for this prop to be enabled "by default". Most text inputs with labels in the platform I've seen do not have bold treatment by default. I'm not really comfortable adding this right now unless we're 100% sure that we're not going to get any developer / product / design pushback on this.

I think we're also unclear on when designers would choose to bold or not bold something, which sort of is another data point that we're not really sure if this should be enabled by default, and it's certainly not used a lot currently, so again I'm just wary of making this a default.

PR #392, opened 2020-07-02
UIP-848: Update react-day-picker

reywright:
## Description
- Update to `react-day-picker` minor broke tests
- Rewrote tests with react-testing-library so we never have to rewrite
- Enabled ability to write async tests
- Allow passing of aria-label to TextInput

Other notes:

- This is a GOOD PR to show how you might rewrite enzyme tests in react-testing-library
- I was able to remove "all" element / class / component name selectors
- Some more advanced typing interaction (backspacing) here as well
- Probably a good "use case" to demo to people (as well as just showing off the power of the library 🎉 )
- Use `screen.debug()` often! (use `DEBUG_PRINT_LIMIT=10000000` env if you need the log to be longer as it will truncate)

No minor bump here. TextInput not accepting an aria-label I consider a bug.

I updated some other tangential packages to make things easier.

## Screenshots
N/A

## Checklist
- [x] All unit tests pass


src/components/DateInput/dateInput.test.js
```diff
@@ -54,88 +56,84 @@ describe('DateInput component', () => {
...
     });
 
-    it('sets DayPicker selected date correctly from input', () => {
-      input.simulate('change', { target: { value: '4/20/2020' } });
-      const dayPickerDate = wrapper.find('DayPicker').prop('month');
-      expect(dayPickerDate.getMonth()).toBe(3);
-      expect(dayPickerDate.getFullYear()).toBe(2020);
+    it('sets DayPicker selected date correctly from input', async () => {
+      await userEvent.type(screen.getByRole('textbox'), '4/20/2020')
+      expect(screen.getByRole('gridcell', { selected: true, name: 'Mon Apr 20 2020' })).toBeTruthy()
```
reywright, line 110: The previous way just finds a prop. This finds the "actual" selection in question by it's role, label text, and then finally that it's selected (all the other options are unselected except this one)

src/components/DateInput/dateInput.test.js
```diff
@@ -54,88 +56,84 @@ describe('DateInput component', () => {
...
+
+      await userEvent.type(screen.getByRole('textbox'), '{selectall}')
+      await userEvent.type(screen.getByRole('textbox'), '{backspace}')
 
       // did the callback fire with null?
       expect(dateChangeFn).toHaveBeenCalledWith(null);
 
       // is the date cleared in the picker?
-      expect(dayPickerSelectedDays).toBe(null);
+      expect(screen.queryByRole('gridcell', { selected: true, name: 'Mon Apr 20 2020' })).toBeNull()
```
reywright, line 127: note that this is using `queryByRole`. `getByRole` fails if it can't find the element, so if you want to check that something doesn't exist, you need to use `queryByRole`.

src/components/DateInput/dateInput.test.js
```diff
@@ -54,88 +56,84 @@ describe('DateInput component', () => {
...
       expect(dateChangeFn).not.toHaveBeenCalled();
-      input.simulate('change', { target: { value: '4/20/2020' } });
+      await userEvent.type(screen.getByRole('textbox'), '4/20/2020')
       expect(dateChangeFn).toHaveBeenCalled();
-      input.simulate('change', { target: { value: '' } });
+
+      expect(screen.getByRole('gridcell', { selected: true, name: 'Mon Apr 20 2020' })).toBeTruthy()
+
+      await userEvent.type(screen.getByRole('textbox'), '{selectall}')
+      await userEvent.type(screen.getByRole('textbox'), '{backspace}')
```
reywright, line 121: https://github.com/testing-library/user-event

This is a separate library that enables a lot of more advanced user interactions (file uploads, selecting / deselecting, hovering, pasting, tabbing, etc.

There's a single command to clear the text as well but I like the clarity of doing it this way.

src/components/DateInput/dateInput.test.js
```diff
@@ -54,88 +56,84 @@ describe('DateInput component', () => {
...
-      expect(dayPickerDate.getMonth()).toBe(3);
-      expect(dayPickerDate.getFullYear()).toBe(2020);
+    it('sets DayPicker selected date correctly from input', async () => {
+      await userEvent.type(screen.getByRole('textbox'), '4/20/2020')
+      expect(screen.getByRole('gridcell', { selected: true, name: 'Mon Apr 20 2020' })).toBeTruthy()
     });
 
-    it('clears selected date when user backspaces out the input', () => {
-      const dayPickerSelectedDays = wrapper.find('DayPicker').prop('selectedDays');
+    it('clears selected date when user backspaces out the input', async () => {
```
reywright, line 113: these tests run async

src/components/DateInput/dateInput.test.js
```diff
@@ -54,88 +56,84 @@ describe('DateInput component', () => {
...
       expect(dateChangeFn).not.toHaveBeenCalled();
-      input.simulate('change', { target: { value: '2/30/2020' } });
+      await userEvent.type(screen.getByRole('textbox'), '2/30/2020')
       expect(dateChangeFn).not.toHaveBeenCalled();
     });
 
     it('calls onDateChange when user selects a day in the picker', () => {
       expect(dateChangeFn).not.toHaveBeenCalled();
-      wrapper.find('Day').at(7).simulate('click');
+      fireEvent.click(screen.getByText('7'));
```
reywright, line 98: This is me being lazy. Technically `screen.getByText` should be your absolute last resort, and almost should never be used to grab something you then want to subsequently click on. I'm probably going to merge "as is" so that this comment / example can stay here for posterity & reference.

src/components/Menu/__snapshots__/menu.test.js.snap
```diff
@@ -47,7 +47,7 @@ exports[`Popover component matches snapshot (JSX items) 1`] = `
             "element": <div
               data-reach-menu-item=""
               data-valuetext="First"
-              id="option-0--menu--5"
+              id="option-0--menu--7"
```
reywright, line 50: I don't know why this is changing, but it doesn't matter because I'm nuking this snapshot the first opportunity I get around to it.

src/components/DateInput/dateInput.test.js
jjniczCB: this is great

src/components/DateInput/dateInput.test.js
jjniczCB: Is this done because there's no clear aria role to use for selection?

donovanm:
Very nice!

kchen-cbi:
Definitely agree this should be pinned for reference, great stuff!

src/components/DateInput/dateInput.test.js
rey-wright: this is sort of silly but to give you the short answer: it shouldn't be done like this and there's probably an accessible hook. I figured it was instructive to show me doing it the wrong way

To give the full answer (only read if you're bored or are really interested):

So when I was trying to test this, `screen.debug()` wasn't outputting the full source. So I sort of "guessed" that the text of 7 would be in the list, and the test passed. Later on when I figured out how to get `screen.debug()` to output the full source, I realized the PR could be used as a tutorial of sorts. Given that, I sort of wanted to leave this here to show people both what you shouldn't do, but also at the same time show this selector off because it's likely people will need it when doing tests on non interactive stuff. My plan is to basically fast follow this PR with a fix specifically to this line 🙃. It's silly reasoning but I figure it's well intentioned and I'm leaving the test much better than what it was.

jjniczCB: just seeing your response now sorry! Thanks for the update, and follow up PR!

PR #401, opened 2020-07-14
UIP-761: replace classnames usage with classcat

rey-wright:
## Description
Classcat is up to 2.5x faster, smaller and the api is "basically" the same. There's almost no reason not to use it!

Note: I "did not" change the code for these files manually (other than one hbs file). I created a codeshift transform and it did all the work 🎉 

## Screenshots
n/a

## Checklist
- [x] All unit tests pass


scripts/bash/transforms/classnames-to-classcat.js
```diff
@@ -0,0 +1,63 @@
...
+          j.importDeclaration(
+            [j.importDefaultSpecifier(j.identifier("cc"))],
+            j.literal("classcat")
+          )
+        );
+      }).toSource({quote: 'single'})
+  
+    const sortArguments = (args) => {
+      return args.sort((a, b) => {
+        if (a.type === "ObjectExpression") {
```
akdetrick, line 44: cool!

src/components/Flex/__snapshots__/flex.test.js.snap
```diff
@@ -10,7 +10,7 @@ exports[`Flex component matches snapshot (default props) 1`] = `
 
 exports[`Flex component matches snapshot (set all props) 1`] = `
 <div
-  className="flex flex--alignEnd flex--column flex--column--reverse flex--justifySpaceBetween flex--noGutters"
+  className="flex--column--reverse flex--justifySpaceBetween flex--noGutters flex flex--alignEnd flex--column"
```
akdetrick, line 13: This is when I like snapshots - we don't need to test classcat, as it's already covered by unit tests. We only need to see that usage has the desired output.

src/components/GapList/index.jsx
```diff
@@ -11,12 +11,12 @@ const GapList = ({ items, xGap = 'm', yGap = 'm', valign = 'top' }) => (
...
       <div
         key={i}
-        className={cx(
+        className={cc([
           'display--inlineBlock',
           `valign--${valign}`,
           `margin--x--${xGap}`,
-          `margin--y--${yGap}`
-        )}
+          `margin--y--${yGap}`,
```
akdetrick, line 18: jscodeshift does a good job with code style - I like that it added trailing commas

rey-wright: This is most likely Prettier potentially. JSCodeshift does do a very good job on formatting, but it's not as smart as Prettier. One of the bad parts of not having prettier in cbi-site yet, is running the transforms on cbi-site introduces formatting errors because jscodeshift can't know the opinionated formatting rules of eslint.

rey-wright:
@akdetrick the cool thing about this transform, is that I ran it "as is" on cbi-site and it worked. hundreds and hundreds of lines changes, I nuked the `classnames` package, installed `classcsat`, and the site seemed fully functional.

I don't really plan on making a MR for that anytime soon (probably smart to just see how it functions within FDS, and also give people a break from conflicts for a bit) but it was interesting information to know.

src/components/Flex/__snapshots__/flex.test.js.snap
rey-wright: hmmmmmmmmm 🙃

PR #406, opened 2020-07-15
use moment to parse incoming strings from  as UTC

akdetrick:
## Description

**The problem**

In chrome, these dates are parsed the following way
```
new Date('2010-1-1'); // jan 1 2020
new Date('2010-01-01); // dec 31 2019
new Date('2010/01/01); // jan 1 2020
new Date('2010/1/1'): // jan 1 2020
```
In firefox, half of the above examples result in 2019 dates. Can you guess which ones? I can't either, because native date parsing in JS is a mess and relying on it for `defaultDate` created a bug in production.

**The solution**
Use moment to parse date strings passed to `defaultDate` instead of using the native date constructor to parse the string.

## Checklist
- [x] All unit tests pass


src/components/DateInput/dateInput.stories.mdx
```diff
@@ -42,14 +42,16 @@ prop.
 ## Default selected date
 
 The input element in `DateInput` will be empty by default.
-Pass an instance of `Date` or a parsable date string (`MM/DD/YYYY`) to the `defaultDate` prop to set a default value.
+Pass an instance of `Date` or a parsable date string to the `defaultDate` prop to set a default value.
+Due to inconsistency across browsers with how native `Date` parsing works, we recommend
+passing a string to `defaultDate`.
```
sfrieson-cbi, line 47: Is it worth modifying this so that the description only talks about string dates, and then add a deprecation notice about date objects (that is, if you actually want to deprecate it)? I'm just think that strting the sentence with "Pass an instance of `Date`..." some might read no further.

src/components/DateInput/index.jsx
```diff
@@ -181,8 +181,9 @@ const DateInput = ({
   maxDate,
   ...rest
 }) => {
-  const defaultDate =
-    typeof defaultDateInput === 'string' ? new Date(defaultDateInput) : defaultDateInput;
+  const defaultDate = moment(defaultDateInput)
```
sfrieson-cbi, line 184: If no `defaultDate` is supplied you'll get '1970-01-01'.

akdetrick: !! you're right!

src/components/DateInput/dateInput.stories.mdx
rey-wright: Removing the deprecation down the line sounds like FDS `v10` energy to me 🙃

src/components/DateInput/dateInput.stories.mdx
```diff
@@ -42,14 +42,14 @@ prop.
 ## Default selected date
 
 The input element in `DateInput` will be empty by default.
-Pass an instance of `Date` or a parsable date string (`MM/DD/YYYY`) to the `defaultDate` prop to set a default value.
+Due to inconsistency across browsers with how native `Date` parsing works, we recommend passing a string to `defaultDate`.
```
rey-wright, line 45: Would we consider getting rid of this sentence completely?

akdetrick: I'd like to keep it in, as the prop types table still declares `Date` as valid for this prop

rey-wright: I understand the information it conveys but from the consumer perspective they don't actually know this accepts a native date object unless we tell them.

akdetrick: We unfortunately do tell them, in the props table, until we remove support in a major version

src/components/DateInput/index.jsx
sfrieson-cbi: Or maybe it's '1969-12-31'. I have no idea anymore...

src/components/DateInput/dateInput.stories.mdx
rey-wright: makes sense 👍

PR #414, opened 2020-07-24
💫 add stopPropagation for popover open

YunfeiYan:
## Description
add stopPropagation for popover open

## Checklist
- [x] All unit tests pass


sfrieson-cbi:
Will this negatively affect click outside functionality if something is already open and they click on this trigger?

rey-wright:
It's a good question. Going to just think out loud here:

- If a modal is open, it takes up the entire screen so it wouldn't matter in that scenario.
- If another popover is open and you click on a new trigger, it wouldn't bubble up to another existing open popover, but that should be fine: you just have another popover render above that one.
- The problem you might run into, is if you have 2 popovers open, then you'd fire the click outside functionality on both of them later on.

Maybe there's a better fix here where we can put the stop propagation inside the trigger?

```
<Popover
  trigger={
    <StopPropagation>
      <Button />
    </StopPropagation>
  }
```

rey-wright:
Essentially the base problem here, is the idea that you have 2 popovers open at the same time, and that `onClickOutside` will indiscriminately close both of them at a later time. We "might" need some sort of manager here if we're going to be rendering multiple popovers at the same time.

rey-wright:
actually, if the outside click is fired "strictly" on mouse down, we might be fine here, I'm doing some research

YunfeiYan:
It's working fine for now. But I don't have 2 popovers open together scenario.

rey-wright:
@sfrieson-cbi @YunfeiYan 

Summary: I think we're fine here.

Details:

I just confirmed that popover dismisses on `mouseDown`. Essentially, as long as any popover currently open is leveraging `mouseDown` vs `click` for `onOutsideClick`, that existing popover will get dismissed before the click functionality fires, because click is mouseDown + mouseUp. I also pushed up an additional change to one of our example stories (stacking story) so that it's easy for us to visualize 2 popovers and how they work together. They seem to be completely functional to me.

PR #439, opened 2020-09-16
148/guidelines/uip 925 add ui copy guidelines

akdetrick:
## Description
Adds UI copy guidelines to storybook

## Screenshots
Screen Shot 2020-09-16 at 7 06 30 PM
Screen Shot 2020-09-16 at 7 06 43 PM

## Checklist
- [x] All unit tests pass


reywright (approved):
My only feedback here is that without a more true graphical representation of what these things would look like in the platform (screenshot for the do and don't), this may be a bit disconnected, and also without unique images to break this up, all these things sort of just blur together and even I have a tough time focusing on each one. I think the content is legit, my only thing is around adding images for that additional context, but I definitely won't block this PR for that.

akdetrick:
I didn't change up the format much in the migration from confluence - we can add in more visual examples later if need be

PR #441, opened 2020-09-18
UIP-936: Add position prop to Tooltip

kchen-cbi:
## Description
In replacing a tooltip in cbi-site with the appropriate FDS component, I noticed that the existing tooltip was intended to be positioned to the right of the trigger. This change adds a `position` prop that is passed through to `Popover` to enable this behavior instead of only positioning under the trigger (`position="bottom"`).

## Checklist
- [x] All unit tests pass


src/components/Tooltip/index.jsx
```diff
@@ -1,24 +1,25 @@
...
-    position="bottom"
+    position={position}
     alignment="center"
     transitionName="GrowFast"
     delay={DELAY_MS}
   >
     <div
       role="tooltip"
-      style={{ maxWidth: maxWidth ? `${maxWidth}px` : '240px' }}
+      style={{ maxWidth: maxWidth ? `${maxWidth}px` : DEFAULT_WIDTH }}
```
reywright, line 22: nice catch

reywright (approved):
Looks good to me, thanks for the PR!

PR #453, opened 2020-10-20
abstract scroll logic + hooks into useGetScrollAttributes reusable hook

reywright:
## Description
I actually might need to use this logic in Rodeo. I'm probably going to add a single hook for it, so I figure I would make this into a single reusable hook in FDS that way it's a simple copy & paste over. I also feel like all the logic state here is relevant to itself only so wanted to create a tighter scope for it to clarify it's relevancy.

## Screenshots
N/A

## Checklist
- [ ] All unit tests pass


src/components/Hscroll/useGetScrollAttributes.js
```diff
@@ -0,0 +1,37 @@
...
+  const [isAtScrollEnd, setIsAtScrollEnd] = useState(false);
+  const contentEl = useRef(null);
+
+  const onScroll = (e) => {
+    const { scrolled, scrollEnd } = getScrollAttributes(e, contentEl.current.scrollWidth);
+    rafSchd(setIsScrolled(scrolled));
+    rafSchd(setIsAtScrollEnd(scrollEnd));
+  };
+
+  return [onScroll, contentEl, isScrolled, isAtScrollEnd];
```
reywright, line 34: @akdetrick 

how do the following sound to you?

- isAtScrollStart
- isAtScrollEnd

src/components/Hscroll/index.jsx
```diff
@@ -33,9 +14,7 @@ const Hscroll = ({
   children,
   ...rest
 }) => {
-  const [isScrolled, setIsScrolled] = useState(false);
-  const [isAtScrollEnd, setIsAtScrollEnd] = useState(false);
-  const contentEl = useRef(null);
+  const [onScroll, contentEl, isScrolled, isAtScrollEnd] = useGetScrollAttributes();
```
akdetrick, line 17: did we want to change `isScrolled` to `isAtScrollStart`? The value would be flipped, but it might be more clear

src/components/Hscroll/useGetScrollAttributes.js
akdetrick: sounds good!

akdetrick (changes requested):
switch `isScrolled` with `isAtScrollStart` (boolean is the opposite, but the concept is easier to grasp)

akdetrick (approved):
🚢

PR #455, opened 2020-10-20
150/component/uip 966 status component updatable refactor

akdetrick:
## Description

Adds `Status` as a standalone notification bubble.
Refactors `Updatable` to use status.
Updates stories for each.

## Screenshots
Screen Shot 2020-10-20 at 5 29 19 PM
Screen Shot 2020-10-20 at 5 29 25 PM
Screen Shot 2020-10-20 at 5 29 57 PM
Screen Shot 2020-10-20 at 5 30 05 PM

## Checklist
- [x] All unit tests pass


reywright (approved):
The only thing I would say re: our conversation yesterday, is to take a look at these tests (the full file): https://github.com/cbinsights/form-design-system/blob/master/src/components/Button/index.test.js

I like that format but if you're not with it, not going to stop this PR from going out.

reywright:
I feel like there "might" be a similar argument here to the argument we make about classnames. We talk a lot about how we don't optimize for typing less with our class name usage. I feel like the only downside to these tests is that you have to type out everything but I love the explicitness of it and I don't mind typing a bit more to ensure someone can't just break the component and regen snapshots to pass the test.

reywright:
Screen Shot 2020-10-21 at 10 59 34 AM

I also sort of wonder about things like this. My current plan as I'm changing Avatar is to remove the product specific bits, for 2 reasons.

1. The FDS library is "supposedly" for people who just want to use it in their projects. I feel like there might be something to only including product specific documentation outside the component documentation.
2. It might make it easier for product people to scan a list of product specific concerns. Also, when we talked about the structure of addding product like documentation, we decided there would be component product documentation for things like dialogs. If we had product documentation for "status" (as updatable is more about positioning) then that would be in line with the dialog documentation and would go in there, outside of this document???

akdetrick:
To address current concerns:

- I will change the name (probably `Count` with types "notification" and "passive")
- Updating tests to RTL following the button example

As for the concern around the table row story - table rows are not necessarily specific to our product in my opinion as they can be found throughout the web. The story does give specific guidance on placement relative to rows according to standards decided by our CBI design team, but I feel that it's similar to the guidance we give on `isDestructive`.

It may be a good idea for me to draft an actual PRD for the style guide so we can debate and align on what kind of documentation goes where. I added the table row story here under the assumption that docs pertaining to a specific component should be colocated with other docs about that component.

PR #465, opened 2020-10-30
UIP-985: migrated tooltip to use Tippy

reywright:
## Description
Couple of notes:
- This implementation does not use headless Tippy, mostly because I don't really see any reason to for generic Tooltips, and non headless will give us some good animation props + themes out of the box we can use. I think headless "might" make more sense for Popover but I'm not even sure about that yet.
- This splits up Tooltip & Popover which to me is the right move as they're different things. If we continue to build tooltip on Popover, anything we want to add on tooltip must be supported on Popover which I don't know is what we want. In any case the code I used to replace this with is super super small.
- Right positioning just works on this implementation
- addition / deletion wise, I only added 3 lines to the line count for the Tooltip file itself.
- Tooltip sanitizes the content by default. We can expose a prop which will allow developers to turn this off.

Breaking changes:
- Text-alignment is removed. I mostly don't see any reason why we would have differing text alignments on a simple message, and it would require additional html vs this implementation where we're sidestepping that altogether.
- Tippy requires that the trigger be an focusable element for accessibility reasons (https://atomiks.github.io/tippyjs/v6/accessibility/#use-natively-focusable-elements) which to me is a good thing. It also triggers the active state when you hover over which is pretty cool.
- I added ref support just to show it off but it might be better if we remove, let someone ask for it, and then we just expose that bit of the Tippy.js that allows for that (this is actually not that straightforward if you try to do it manually with Popper.js from my experience, so it's a good win).

## Checklist
- [ ] All unit tests pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -156,8 +156,9 @@
   },
   "version": "4.0.0",
   "dependencies": {
+    "@popperjs/core": "^2.5.4",
     "@reach/menu-button": "^0.10.5",
-    "@popperjs/core": "^2.4.1",
+    "@tippyjs/react": "^4.2.0",
     "classcat": "^4.1.0",
     "focus-trap-react": "^6.0.0",
     "raf-schd": "^4",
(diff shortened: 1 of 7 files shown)


reywright:
Closing for now until we start nearing Q1.

PR #467, opened 2020-11-02
UIP-987: adding ZeroState to FDS

reywright:
## Description
ZeroState is being migrated to FDS and will be removed in cbi-site Storybook going forward.

Thanks @tonysight for creating this component!

## Screenshots
Screen Shot 2020-11-02 at 11 24 06 AM

## Checklist
- [x] All unit tests pass


src/components/ZeroState/index.spec.js
```diff
@@ -0,0 +1,22 @@
...
+    });
+  });
+
+  describe('if Icon is provided', () => {
+    it('should render icon', () => {
+      render(<ZeroState Icon={TableIcon} />);
+      expect(screen.getByRole('img')).toBeTruthy();
+    });
+  });
+});
```
reywright, line 22: I'm not going to test min-height as I don't think it's critical to the test (and should merely be visual regression snapshot tested) but I'm open to pushback.

src/components/style/zeroState.css
```diff
@@ -0,0 +1,3 @@
+.pointer-events-none {
+  pointer-events: none;
+}
```
reywright, line 3: this class was originally `.zeroState`. Given that this looks like it could easily be a utility class, what I like to do is just make it a utility class and then it can just be moved into a base class folder easily later (it sort of greases the wheel for adding utility classes which is why I do this, but don't move it immediately to blow out the scope of the PR)

src/components/ZeroState/index.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import FDS from 'lib/dictionary/js/styleConstants';
+
+const ZeroState = ({ Icon, label, height, ...props }) => (
```
jjniczCB, line 5: Since this component is primarily rendered as a replacement for other content, could that logic live within ZeroState?

something like the below where children is the component normally rendered when results are available
```
return props.resultsCount ? children : <article
    className="alignChild--center--center pointer-events-none"
    style={{ minHeight: height }}
  >
    <section {...props}>
      {Icon && (
        <div className="alignChild--center--center" role="img">
          <Icon size="xl" color={FDS.FONT_COLOR_HINT} />
        </div>
      )}
      <p className="color--hint padding--top--half align--center">{label}</p>
    </section>
  </article>
```

jjniczCB: Thinking of it more, this may be a bit too restrictive to the usage of ZeroState and probably would be better suited for some cbi-site helper functional component

src/components/style/zeroState.css
reywright: decided in meeting: aligned, decide which category

akdetrick: we can replace this custom style with the utility class as a separate PR if you want - just file a follow up ticket before merging

akdetrick (approved):

src/components/ZeroState/index.js
```diff
@@ -0,0 +1,43 @@
...
+  <article
+    className="alignChild--center--center pointer-events-none"
+    style={{ minHeight: height }}
```
akdetrick, line 8: should `height` have a default of `100%` to fill the parent container by default?

reywright: I'm hesitant to do that unless you're 100% confident that there won't be regressions. I think it's safer for these cbi-site storybook --> FDS migrations to be mostly identical in display / functionality as it sort of muddies the waters and makes a simple import switch carry a bit more risk. I'd rather introduce those sort of changes either before migration on cbi-site, or post a more vanilla migration.

That all being said the change itself makes sense.

PR #476, opened 2020-11-06
fix white border on  variant Indicator

akdetrick:
Patch to fix this issue with the white border:

image


reywright:
I have a slight question about the approach here.

So by disabling the border width on one of the themes of the button, we're changing the dimensions of it, which isn't a big deal.



In general how I handle this with FDS Buttons is I add a border width to all buttons, and for buttons that don't need it, I make the border transparent, so that the border itself doesn't change the dimensions of 2 identical buttons which are the same outside of just the theme. What do you think of that?

akdetrick:
I went with removing the border because it could be done via utility classes. I wasn't that concerned with the 2px bump because the element will center itself within a parent container regardless

reywright:
> I went with removing the border because it could be done via utility classes. I wasn't that concerned with the 2px bump because the element will center itself within a parent container regardless

Would it center itself in this situation as well?

image

reywright (approved):
Couple of concerns but not a super big deal.

PR #477, opened 2020-11-10
Allows events for checkbox with supplied id

sfrieson-cbi:
## Description
- Uses the user-supplied `id` if there is one
- avoids component-supplied props being overwritten by user-supplied rest props.
- converts `enzyme` tests to `rtl`

When developing a component that needed to render its own label due to needing markup in the label I used a combination of `id="my-own-id"` and `showLabel={false}`. Due to the fact that the rest props were applied last, my supplied `id` was overwriting the component-supplied `id`. This led to my custom label working properly, while the checkbox did not. This was because the checkbox was made of SVGs inside a label with an `htmlFor` linked to an `id` which wasn't being applied.

This same overwriting is possibly happening for any other prop there (`type`, `className`, etc.), so because of that this could arguably be seen as either a bug fix or a breaking change. A non-breaking change alternative would be to replace the rest props where they were.

## Checklist
- [x] All unit tests pass


src/components/Checkbox/index.test.js
```diff
@@ -1,48 +1,52 @@
...
+    render(
+      <>
+        <Checkbox id="checkbox-id" label="Component label" showLabel={false} name="multiple labels" />
+        <label htmlFor="checkbox-id">User label</label>
+      </>
+    );
+
+    expect(screen.getByLabelText('Component label')).toBeTruthy();
+    expect(screen.getByLabelText('User label')).toBeTruthy();
+  });
```
reywright, line 51: Thanks for rewriting all of this to react-testing-library ❤️

src/components/Checkbox/index.test.js
```diff
@@ -1,48 +1,52 @@
...
-    const wrapper = shallow(<Checkbox disabled label="Snapshot Label" name="snapshot_checkbox" />);
-    expect(wrapper).toMatchSnapshot();
+  it('passes the `disabled` prop', () => {
+    const { container } = render(<Checkbox label="change callback" disabled name="disabled-test" />);
+    const input = screen.getByLabelText('change callback');
+    expect(input.disabled).toBeTruthy();
+
+    fireEvent.click(input);
+    expect(input.checked).toBeTruthy();
+    expect(container.querySelector('.fdsCheckable')).toHaveClass('fdsCheckable--disabled')
```
reywright, line 15:
```suggestion
    expect(screen.getByRole('checkbox')).toHaveAttribute('disabled'
```

reywright: What do you think of this suggestion?

reywright (approved):
Looks good, fairly straightforward. The only thing I would recommend is getting feedback from Adam here as well, "specifically" around the idea of supporting the disabling of the label, not necessarily to block this PR, but to have a good use case for the ensuing discussion around a Checkbox refactor we have planned & scheduled. Adam has a strong vision around there could be a good opportunity for you both to chat about it.

src/components/Checkbox/index.test.js
sfrieson-cbi: I was following the styling already existent in other tests for the class check. I think your suggestion is a slightly different expectation that has value as well.

sfrieson-cbi:
For the label, it has been disable-able with `showLabel={false}`. The issue is that the user may or may not supply an alternative label and they shouldn't necessarily have to do so. Adding the `aria-label`, is ensuring that the checkbox still has a label even when it is hidden.

sfrieson-cbi:
Hey @akdetrick, Rey mentioned that you might have opinions related to this MR. Do you want to look it over?

reywright:
@sfrieson-cbi talked with @akdetrick . I think this is free to go forward... we'll most "likely" support disabling the label in the future but if we don't, it'll just be a breaking change we'll adjust.

PR #479, opened 2020-11-18
update ADR template to match standard

akdetrick:
Updates ADR template to match the company standard (found in the gitlab main ADR repo)


reywright (approved):
Any concern about these 2 getting out of sync (vs just linking to that ADR directly)?

akdetrick:
I'm not that concerned - if there's a significant change, we can always update the template again manually, but I don't expect that to happen now that we've settled on a template

PR #482, opened 2020-11-20
Changes file extension

sfrieson-cbi:
## Description
Changes the file extension for the `ZeroState` component. I tried to import it as `import { ZeroState } from '@cbinsights/fds/lib/components';` and it resulted in the common bad import error:
```
Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
```

In checking the source, I realized it was a `.js` file and not `.jsx` so I hypothesized that this might fix it.

## Checklist
- [x] All unit tests pass


akdetrick (approved):
thanks, let's give it a shot

PR #485, opened 2020-11-25
added useScrollAttributes hook

reywright:
## Description
Added / exposed useScrollAttributes hook

- First FDS hook
- Removed the scroll callback as we already have a ref, which I've now applied listeners to directly
- I also made it so that the hook "accepts" a ref. This is generally going to be a better strategy if I need to pass a ref through multiple hooks or if I already have a hook I want to use.
- I could not add fade classes (for that extra bit of visual flair for the example) as those need to be created from scratch using `:after` similarly to what `Hscroll` does. That'll come after this PR.

## Screenshots
Screen Shot 2020-11-25 at 11 16 55 AM

## Checklist
- [x] All unit tests pass


src/hooks/useScrollAttributes/index.js
```diff
@@ -20,21 +20,24 @@ export const getScrollAttributes = (e, contentScrollWidth) => {
...
-  const contentEl = useRef(null);
 
   const onScroll = (e) => {
-    const { scrollStart, scrollEnd } = getScrollAttributes(
-      e,
-      contentEl.current.scrollWidth
-    );
+    const { scrollStart, scrollEnd } = getScrollAttributes(e, ref.current.scrollWidth);
     rafSchd(setIsAtScrollStart(scrollStart));
     rafSchd(setIsAtScrollEnd(scrollEnd));
```
ahmedhashim-cbi, line 30: Would these two be better as a single state update to an object with these 2 bools instead of 2 useState calls? My thinking is unless rafSchd can ensure batching those updates together during reconciliation, React might unmount this hook after the first & raise a `Can't perform a React state update on an unmounted component` warning. 

Something like this can help to prevent that:
```
[isAtScrollPositions, setIsAtScrollPositions] = useState({ start: true, end: false });
...
const { scrollStart: start, scrollEnd: end } = getScrollAttributes(e, ref.current.scrollWidth);
rafSchd(setIsAtScrollEnd({ start, end }));
```

Another alternative would be to use `useReducer` instead of `useState` entirely (but doesn't seem necessary in this case).

ahmedhashim-cbi:
Nice work - this looks handy!

src/hooks/useScrollAttributes/index.js
reywright: Thanks! (talked offline regarding this). I "might" rather see this break on us as this would create some bigger pattern ramifications across the board for cbi-site if this is actually an anti-pattern 😅

akdetrick (approved):

PR #499, opened 2020-12-09
[V10] [4] UIP-974: remove old margin

reywright:
## Description
Removed double and half support from spacing classes

## Checklist
- [ ] Docs have been rebuilt (`yarn build:full`)
- [ ] All unit tests pass
- [ ] Snapshots have been updated
- [ ] No lint errors or warnings have been introduced
- [ ] **[Version bumped](https://github.com/cbinsights/form-design-system#updating-version-number) if appropriate**

PR #500, opened 2020-12-09
[V10] [5] UIP-968 add margin support to button

reywright:
## Description
Added FDS margin support to FDS Button just to try it out. If it's well received, it's incredibly easy to add this to whatever component we need to add it to.

## Checklist
- [ ] All unit tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,16 @@ This is an example of a brief overview of the _Major_ or _Minor_ version changes
 #### Component reorganization
 - bgFade classes are removed
 
+#### Double and Half spacing removed
+- classes, css variables, and js variables all updated
+- upgrade guide: 
+  - `half` (8px) maps to `s` (8px)
+  - `double` (32px) maps to `xl` (32px)
+- Examples:
+  - CSS Variable: { margin: var(--space-half) } becomes { margin: var(--space-half) }
+  - JS Variable: FDS.SPACE_DOUBLE becomes FDS.SPACE_XL 
+  - classes: .margin--top--half .becomes margin--top--s
+
 #### Removed components
 - `HScroll` (use fade classes + useScrollAttributes hook)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@ This is an example of a brief overview of the _Major_ or _Minor_ version changes
   - `half` (8px) maps to `s` (8px)
   - `double` (32px) maps to `xl` (32px)
 - Examples:
-  - CSS Variable: { margin: var(--space-half) } becomes { margin: var(--space-half) }
+  - CSS Variable: { margin: var(--space-half) } becomes { margin: var(--space-s) }
   - JS Variable: FDS.SPACE_DOUBLE becomes FDS.SPACE_XL 
   - classes: .margin--top--half .becomes margin--top--s
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,8 @@ This is an example of a brief overview of the _Major_ or _Minor_ version changes
 
 ### Changes / Additions
 
--  Fade classes added, which work via pseudo elements and support using multiple at a time
+- Fade classes added, which work via pseudo elements and support using multiple at a time
+- FDS Button now has support for adding margins.
 
 ### **BREAKING CHANGES**
 
--- a/src/components/Button/index.jsx
+++ b/src/components/Button/index.jsx
@@ -73,6 +73,7 @@ const Button = forwardRef(
             'fdsButton--disabled': disabled,
             'fdsButton--isFullWidth': isFullWidth,
             'fdsButton--isActive': isActive && !disabled,
+            [`margin--${margin}`]: margin,
             breakoutLink: isBreakoutLink,
           },
           'fdsButton',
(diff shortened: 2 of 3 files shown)


src/components/Button/index.jsx
```diff
@@ -83,6 +84,7 @@ const Button = forwardRef(
           'transition--default',
           `fdsButton--${size}`,
           `fdsButton--${theme}`,
+          `margin--${margin}`,
```
reywright, line 87: I either need to do this:

```
...(cond ? ['a'] : [])
```

or create some memoized function and execute it here. Basically this will leave `margin--undefined` on any button that isn't using margin.

src/components/Button/index.jsx
```diff
@@ -150,6 +152,50 @@ Button.propTypes = {
   Link: PropTypes.func,
   /** Extend click radius of button to nearest relative parent */
   isBreakoutLink: PropTypes.bool,
+  /** Allows specifying a margin class value. all--y, top--xl, etc */
```
reywright, line 155: mistake, this should be `all-y` and `top-xl`

CHANGELOG.md
```diff
@@ -27,6 +27,31 @@ This is an example of a brief overview of the _Major_ or _Minor_ version changes
...
+#### Component reorganization
+- bgFade classes are removed
+
+#### Double and Half spacing removed
+- classes, css variables, and js variables all updated
+- upgrade guide: 
+  - `half` (8px) maps to `s` (8px)
+  - `double` (32px) maps to `xl` (32px)
+- Examples:
+  - CSS Variable: { margin: var(--space-half) } becomes { margin: var(--space-half) }
```
jjniczCB, line 48: think this should be { margin: var(--space-s) }

src/components/Button/index.jsx
```diff
@@ -150,6 +152,50 @@ Button.propTypes = {
...
+      m: true,
+      l: true,
+      xl: true,
+    };
+
+    if (marginValue) {
+      const split = marginValue.split('-');
+      // If the first part of the split class is not a direction, it's an invalid value.
+      if (!directions[split[0]]) {
+        return new Error(`Invalid margin (no valid direction detected)`);
```
jjniczCB, line 183: TIL you can validate props like this, very cool

src/components/Button/index.jsx
```diff
@@ -83,6 +84,7 @@ const Button = forwardRef(
           'transition--default',
           `fdsButton--${size}`,
           `fdsButton--${theme}`,
+          ...(margin ? [`margin--${margin}`] : []),
```
jjniczCB, line 87: [could] move this to line 76 in the object: 
 ```
breakoutLink: isBreakoutLink,
`margin--${margin}`: margin
```

File: CHANGELOG.md (hunk 3 above)
Comment (reywright) [reply CHANGELOG.md:48]: lol good catch, thanks.

File: src/components/Button/index.jsx (hunk 5 above)
Comment (reywright) [reply src/components/Button/index.jsx]: ah yeah good point! Done.

File: CHANGELOG.md (hunk 3 above)
Comment (reywright) [reply CHANGELOG.md]: fixed

File: src/components/Button/index.jsx (hunk 2 above)
Comment (reywright) [reply src/components/Button/index.jsx:155]: actually I had it right. Barnacles...

File: src/components/Button/index.jsx [hunk 6]
```diff
@@ -150,6 +152,50 @@ Button.propTypes = {
...
+      const split = marginValue.split('--');
+      // If the first part of the split class is not a direction, it's an invalid value.
+      if (!directions[split[0]]) {
+        return new Error(`Invalid margin (no valid direction detected)`);
+      }
+      // If there happens to be an amount (which is optional)
+      if (split[1]) {
+        // If a standard amount isn't detected, the class is invalid
+        if (!amounts[split[1]]) {
+          return new Error(`Invalid margin value (no valid value detected)`);
```
akdetrick, line 189: In CSS classes, the amount is optional - we always default to the `m` value.
Should we also allow this in the component?

```
<Button margin="x" />
```

src/components/Button/index.jsx
```diff
@@ -150,6 +152,50 @@ Button.propTypes = {
...
+        }
+      }
+      // There should only be a single double hyphen in the value they pass else it's wrong
+      if (split[2]) {
+        return new Error(`Invalid margin value (should only be a single hyphen)`);
```
akdetrick, line 194:
```suggestion
        return new Error(`Found more than one double hyphen ('--') in margin prop`);
```

src/components/Button/index.test.js
```diff
@@ -43,13 +43,15 @@ describe('Button component', () => {
         disabled	
         isFullWidth	
         isBreakoutLink
+        margin="all-s"
```
akdetrick, line 46: Shouldn't this error with only a single hyphen?

akdetrick (changes requested):
A few questions about the prop validation of `margin`

akdetrick:
Contains unmerged commits from `154/UIP-974-remove-old-margin`; needs to wait on #499 to merge

src/components/Button/index.jsx
reywright: yep that's allowed. `if (!amounts[split[1]]) {` is only run if an amount exists (`split[1]`). If it doesn't , it just falls through.

src/components/Button/index.test.js
reywright: yup you're right :)

src/components/Button/index.jsx
reywright: I broke these out into variables with appropriate names so that the code is easer to understand.

akdetrick (approved):

PR #504, opened 2020-12-10
[V10] [6] UIP-985: tippy tooltip

reywright:
## Description
- Tooltip engine has been changed to tippy.
- API is exactly the same.
- the only breaking change is that the trigger is now "required" to be focusable.
- FDS tooltip styling is intact.

## Screenshots
Screen Shot 2020-12-10 at 11 53 01 AM

## Checklist
- [ ] All unit tests pass


src/components/Tooltip/index.test.js
```diff
@@ -13,28 +12,4 @@ describe('Tooltip component', () => {
...
-    const evil = 666;
-    const tooltipMaxWidth = shallow(
-      <Tooltip
-        trigger={<Trigger />}
-        message="tooltip with evil max width"
-        maxWidth={evil}
-      />
-    ).find(SELECTOR_TOOLTIP).prop('style').maxWidth;
-
-    expect(tooltipMaxWidth).toBe(`${evil}px`);
```
reywright, line 37: max width is non critical IMO to the core of the component which is why I've axed this for now (and it's the sort of thing that would be better to visual snapshot than sniff computed widths and stuff)

src/components/Tooltip/index.jsx
```diff
@@ -1,58 +1,59 @@
...
   maxWidth: PropTypes.number,
 
-  /**
-   * Position preference of tooltip.
-   * `top` for example, will place the tooltip above the trigger.
-   */
-  position: PropTypes.oneOf(VALID_POSITIONS),
-
+  /** Position preference of tooltip. */
+  position: PropTypes.oneOf([
```
akdetrick, line 40: At first I wondered if we should only include centered positioning options (left, right, bottom, left), but I can't think of a good reason not to include all positioning options.

We might want to put this array somewhere shared for Popover, but we can decide on that as you work with the Popover conversion

src/components/style/index.css
```diff
@@ -39,3 +38,8 @@
 @import "./toaster.css";
 @import "./dialog.css";
 @import "./menu.css";
+/* Reach framework styles */
```
akdetrick, line 41: I had put third party styles higher in the cascade to guarantee any conflicting rules in FDS would override. What do you think about keeping both reach and tippy styles above specific component styles?

CHANGELOG.md
```diff
@@ -35,6 +35,9 @@ This is an example of a brief overview of the _Major_ or _Minor_ version changes
 
 ### **BREAKING CHANGES**
 
+#### Tooltip component
```
akdetrick, line 38: We should also add a note for the non-breaking change of extra positioning options for `Tooltip`

src/components/Tooltip/__snapshots__/index.test.js.snap
```diff
@@ -5,82 +5,55 @@ exports[`Tooltip component matches snapshot (uses the right classes) 1`] = `
...
-        }
-      >
-        <Trigger />
-      </span>
+  <ForwardRef(TippyWrapper)
+    appendTo={<body />}
+    arrow={false}
+    className="bgColor--charcoal inverted align--center padding--top--half padding--bottom--half padding--left padding--right elevation--2 rounded--all fontSize--s fontWeight--bold"
+    content="lol"
+    delay={
```
akdetrick, line 13: One requested feature we had for Tooltip previously was a delay.

I feel pretty strongly that the delay should **not** be configurable; it should be an on or off option so that we can make hover delays consistent between components.

Would it be difficult to add a `hasDelay` prop that defaults to true?

src/components/style/index.css
```diff
@@ -39,3 +38,8 @@
...
+@import "@reach/menu-button/styles.css";
+/* Tippy framework styles */
+@import "tippy.js/dist/tippy.css";
+@import "./tooltip.css";
```
akdetrick, line 45: I think we have two options for organizing tippy themes:

1. Dedicated theme dir: `src/components/style/tippy/<theme_name>.css`
2. Inlined in component css: `src/components/style/<name_of_component_that_uses_tippy>.css`
 
I like option 1 because it's clear that the CSS is targeting tippy-specific selectors and solidifies the relationship of that CSS to the third party theming conventions.

I like option 2 because if a developer thinks "i need to change tooltip styles", they don't have to do extra digging to find the tippy theme.

Interested in your thoughts here.

akdetrick (changes requested):
Couple of questions/comments to resolve, but seems pretty solid

src/components/Tooltip/index.jsx
reywright: I mean less options is better if we know they don't need them. That being said, that would be easy to go through and do a breaking change at a later date, if we discover what the "UX" is from our designers or if we see common patterns.

src/components/Tooltip/__snapshots__/index.test.js.snap
reywright: What do you think about having that be a separate task / ticket as a followup? It wouldn't be difficult in either case.

I'm with you on not having delay configurable and just relying on a boolean in the facade.

src/components/style/index.css
reywright: Hmmmm yeah this is damn if you do damn if you don't. If we choose option 1, then people who go to edit the "tooltip css" are going to be confused. If we choose option 2, there's a liability that someone can create an identical theme which can mess with another.

The big issue here is that we're not just applying a theme, we're applying a "type".

I'm trying to think if there is a "combination" route we can go here. like we have a place where we put the themes, but then we import the themes into a specific component. That way you get all the separated themes in one place, you still have the component specific css, which is helpful for people to find their way.

src/components/Tooltip/index.jsx
akdetrick: I'm good with leaving all these in for now, agree with your point

src/components/Tooltip/__snapshots__/index.test.js.snap
akdetrick: resolving with follow up ticket https://cbinsights.atlassian.net/browse/UIP-1008

src/components/style/index.css
akdetrick: I like the idea of creating a top-level dir for tippy themes `components/style/tippy/<name>-theme.css` and importing them directly in component CSS files instead of into the index CSS

reywright: do you mind if I set up that structure in the v10 popover ticket?

akdetrick (approved):

PR #505, opened 2020-12-10
[V10] [7] UIP-1005: badge --> indicator / indicator --> badge

reywright:
## Description
* Badge is renamed to be Indicator
* Indicator is renamed to be badge

## Screenshots
Screen Shot 2020-12-10 at 12 23 47 PM
Screen Shot 2020-12-10 at 12 23 51 PM

## Checklist
N/A


src/components/Indicator/index.jsx
```diff
@@ -2,41 +2,22 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import cc from 'classcat';
 
-export const VALID_TYPES = ['update', 'count'];
+export const SIZES = ['xs', 's'];
```
akdetrick, line 5: This could be cleaner as a `git mv` - I'm assuming there aren't actual code changes being made with this branch...

akdetrick:
Contains unmerged commits from `151/UIP-985-tippy-tooltip`; needs to wait on #504 to merge

src/components/Indicator/index.jsx
reywright: discussed on zoom, going to just merge in as is for now but I'll try to avoid this in the future.

PR #525, opened 2020-12-21
[V10] UIP-906: popover tippy

reywright:
## Description
Tippy engine!

## Screenshots
Screen Shot 2020-12-21 at 6 45 38 PM

## Checklist
- [ ] All unit tests pass


src/components/style/popover.css
```diff
@@ -0,0 +1,2 @@
+@import "./tippy/popover-default-theme.css";
+@import "./tippy/popover-unthemed-theme.css";
```
reywright, line 2: Did exactly what we discussed: created themes, and then component css imports the theme(s) it needs.

src/components/style/tippy/popover-default-theme.css
```diff
@@ -0,0 +1,15 @@
+.tippy-box[data-theme~="popover-default"] {
+  color: initial;
+  background-color: var(--color-white);
+  border-radius: var(--border-radius-default);
+  box-shadow: 0 4px 5px 0 rgba(0, 0, 0, 0.14),
+    0 1px 10px 0 rgba(0, 0, 0, 0.12),
+    0 2px 4px -1px rgba(0, 0, 0, 0.2);
+  overflow: hidden;
```
reywright, line 8: One thing to discuss: I sort of "don't like" using theme props to conditionally add classnames... I just think it's less flexible and a lot harder to grasp. Because there are more than 1 theme in tooltip, I opted to use classes, but I opted to use css for Popover.

Now given that, I opted to inline the box-shadow of the `elevation--2` class directly. One of my 1:1 items is that I don't really see a reason we can't encapsulate these into variables?

src/components/style/tippy/popover-unthemed-theme.css
```diff
@@ -0,0 +1,11 @@
...
+  background: transparent;
+  color: initial;
+  border-radius: initial;
+  font-size: unset;
+  line-height: initial;
+}
+
+.tippy-box[data-theme~="popover-unthemed"] .tippy-content {
+  padding: 0;
+}
```
reywright, line 11: all the unthemed theme does is zero out all the styles tippy provides.

src/components/Tooltip/index.jsx
```diff
@@ -19,8 +19,8 @@ const Tooltip = ({
     placement={position}
     zIndex={FDS.ZINDEX_MODAL}
     moveTransition="transform 0.2s ease-out"
-    appendTo={document.body}
-    className={`bgColor--charcoal inverted align--${textAlign} padding--top--half padding--bottom--half padding--left padding--right elevation--2 rounded--all fontSize--s fontWeight--bold`}
+    appendTo={() => document.body}
+    className={`bgColor--charcoal inverted align--${textAlign} padding--y--s  padding--x elevation--2 rounded--all fontSize--s fontWeight--bold`}
```
reywright, line 23: not sure how the `half` padding snuck in here, but just fixing it here.

src/components/Popover/index.jsx
```diff
@@ -1,300 +1,57 @@
...
-   * events from user interaction.
-   */
-  interactionMode: PropTypes.oneOf(VALID_INTERACTION_MODES),
-
-  /** Controlls active state of popover when in fully controlled interaction mode */
+  appendTo: PropTypes.oneOfType([PropTypes.element, PropTypes.string]),
+  children: PropTypes.node,
+  /** Element that popover is positioned on */
+  triggerElement: PropTypes.node,
+  placement: PropTypes.oneOf(['auto', 'top', 'right', 'bottom', 'bottom-start', 'left']),
```
reywright, line 50: it may be worth looking at the current placement for popover and tooltip and seeing if we want to align them for whatever reason. The only reason they can't be aligned now is popover supports `bottom-start` which we know is the default way we want popovers rendered. That being said, I don't want to seek consistency with Tooltip just for DRY sake... if we don't ever want to position tooltip with a bottom-start, I think these should stay separate and that it's totally fine to keep separate.

src/components/Popover/index.jsx
```diff
@@ -1,300 +1,57 @@
...
+  children,
+  isOpen,
+  placement = 'bottom-start',
+  trigger = 'click',
+  triggerElement,
+  onClickOutside,
+  theme = 'default',
+  appendTo = () => document.body,
+}) => {
+  const hideOnEsc = useMemo(() => hideOnEscFunc(onClickOutside), [onClickOutside]);
```
reywright, line 17: this is a little tricky but essentially if we pass an onClickOutside here, it'll also get invoked on esc. I'll probably change the prop name to onUserDismiss again as this makes no sense semantically right now.

src/components/DateInput/DateInputPopover/index.jsx
```diff
@@ -0,0 +1,300 @@
...
+
+export const VALID_POSITIONS = ['auto', 'top', 'right', 'bottom', 'left'];
+export const VALID_ALIGNMENTS = ['start', 'end', 'center'];
+export const VALID_INTERACTION_MODES = ['hover', 'click', 'controlled'];
+
+/**
+ * @param {Object} props react props
+ * @returns {ReactElement}
+ */
+const Popover = React.forwardRef(
```
reywright, line 19: This Popover is now `DateInputPopover` as that is the only thing in our codebase that is using it. When you start working on a breaking change for this component, it would be very easy to pull out this popover and just give DateInput it's own simple tippy instance.

src/components/DateInput/DateInputPopover/index.jsx
```diff
@@ -0,0 +1,300 @@
...
+      setIsActive(interactionMode === 'controlled' && isOpen);
+    }, [interactionMode, isOpen]);
+
+    /**
+     * Called when user takes an action to dismiss the popover
+     * @param {Event} e DOMEvent
+     */
+    const handleUserDismiss = (e) => {
+      onUserDismiss(e);
+      if (interactionMode !== 'controlled') {
```
akdetrick, line 58: If `DateInput` is the only thing that uses this, do we need to support multiple interaction modes?

src/components/Popover/hideOnEsc.js
```diff
@@ -0,0 +1,27 @@
+const hideOnEscFunc = (callback) => ({
+  name: 'hideOnEsc',
+  defaultValue: true,
+  fn(props) {
+    console.log('hi adam', props);
```
akdetrick, line 5: ✂️

src/components/Popover/index.jsx
```diff
@@ -1,300 +1,57 @@
...
-
-  /** When popover opens, this element should freeze scrolling. When the popover closes, the scrollability of this element should resume. */
-  disableScrollRef: validRef,
-
-  /** When the Popover detects scroll events from this ref, the popover should close. */
-  closeOnScrollRef: validRef,
-
-  /** Ref to forward to the content container of the popover */
-  ref: validRef,
+  onClickOutside: PropTypes.func,
```
akdetrick, line 52: agreed this should be `onUserDismiss`

src/components/Popover/index.jsx
```diff
@@ -1,300 +1,57 @@
...
+  trigger: PropTypes.oneOf(['click', 'mouseenter focus']),
```
akdetrick, line 53: is `mouseenter focus` supposed to be a single value?

src/components/style/tippy/popover-default-theme.css
reywright: We discussed offline and decided to skip for now?

src/components/Popover/index.jsx
akdetrick: also, how does someone create a controlled popover? Is there a `trigger` value in tippy that allows something like the below example?

`<Popover isOpen={isOpenFromOuterState} onUserDismiss={() => {setIsOpenFromOuterState(false)}} ... />`

src/components/DateInput/DateInputPopover/index.jsx
reywright: I sort of don't see DateInputPopover as a component that will stick around. Me changing the name of the current Popover to DateInputPopover allows DateInput to still work, but the idea is that this component is on life support and would be trashed the MOMENT we get a chance, so I don't see it necessary to improve the component / remove cruft / removed unused props. That's just my perspective though, open to another viewpoint on it.

src/components/Popover/index.jsx
```diff
@@ -1,300 +1,57 @@
...
-  PropTypes.func,
-  PropTypes.shape({ current: PropTypes.any }),
-]);
+        {triggerElement}
+      </Tippy>
+    </>
+  );
+};
 
 Popover.propTypes = {
```
akdetrick, line 45: props need jsDoc comments once we solidify the API

src/components/Popover/hideOnEsc.js
reywright: lol 👍

akdetrick (dismissed):
Some questions to resolve before finishing and merging this

src/components/Popover/index.jsx
reywright: Yup 😁 

Screen Shot 2020-12-29 at 4 03 30 PM

src/components/DateInput/DateInputPopover/index.jsx
akdetrick: "life support" works for me!

src/components/Popover/index.jsx
akdetrick: Nice - so now I'm wondering if we should tuck these options behind our facade component...

`triggerType: PropTypes.oneOf(['manual', 'hover', 'click'])`

In this case, "hover" would map to `mouseenter focus`.

Either way, I think we need to add `manual` as a valid prop value here

reywright: Makes sense. I sort of wonder if adding `isOpen` makes it manual trigger automatically (as the above is tippy, but we're dealing with the react implementation): I'll research and let you know, but definitely agree with the facade.

reywright:
```
<Popover
  appendTo={() => {}}
  isOpen
  onClickOutside={function noRefCheck() {}}
  placement="bottom-start"
  theme="default"
  trigger="click"
  triggerElement={<button onClick={function noRefCheck() {}}>Open popover</button>}
>
  <div className="padding--all">
    <h3 className="type--head3">
      Look at me
    </h3>
    <p>
      <em>
        i am the popover
      </em>
    </p>
  </div>
</Popover>
```

I changed the story so that it's more apparent, but you can control `isOpen` which makes the manual trigger irrelevant. I'm assuming that's for non React implementations. I am going to add the `hover` facade though.

src/components/Popover/index.stories.js
```diff
@@ -1,170 +1,56 @@
...
-          alignment="start"
-          distance={8}
-          closeOnScrollRef={containerEl}
+    <Popover
+      {...args}
+      onClickOutside={() => setIsOpen(false)}
+      triggerElement={
+        <button
+          onClick={() => {
+            setIsOpen(!isOpen);
```
akdetrick, line 35: We could investigate as a follow up, but storybook CSF seems to be doing some "interesting" things with this code example:

Screen Shot 2020-12-31 at 12 25 31 PM

src/components/Popover/index.stories.js
```diff
@@ -1,170 +1,56 @@
...
-        <Popover
-          {...args}
-          trigger={<Button theme="outlined" label="Open popover" />}
-          position="bottom"
-          alignment="start"
-          distance={8}
-          closeOnScrollRef={containerEl}
+    <Popover
+      {...args}
+      onClickOutside={() => setIsOpen(false)}
```
akdetrick, line 31: should be `onUserDismiss`, right?

src/components/Popover/index.jsx
```diff
@@ -1,300 +1,66 @@
...
-  /** When popover opens, this element should freeze scrolling. When the popover closes, the scrollability of this element should resume. */
-  disableScrollRef: validRef,
-
-  /** When the Popover detects scroll events from this ref, the popover should close. */
-  closeOnScrollRef: validRef,
-
-  /** Ref to forward to the content container of the popover */
-  ref: validRef,
+  /** What mode of interaction triggers the popover */
+  trigger: PropTypes.oneOf(['click', 'hover']),
```
akdetrick, line 57: In the current implementation, `trigger` refers to the trigger element. I like the `triggerElement` prop name for the tippy popover because it's explicit.

How would you feel about making this name explicit as well (`triggerType`)?

src/components/Popover/index.jsx
```diff
@@ -1,300 +1,66 @@
...
+  const computedTrigger = triggerHelper(trigger, isOpen);
+  return (
+    <>
+      <Tippy
+        animation={false}
+        arrow={false}
+        visible={isOpen}
+        theme={`popover-${theme}`}
+        trigger={computedTrigger}
+        placement={placement}
```
akdetrick, line 28: Looks like we may want to go ahead and implement Tippy's `offset` prop, with a public prop of `distance`. 

The `distance` prop is being used in a few places in `cbi-site`

src/components/Popover/index.stories.js
reywright: I think this is mostly storybook limitations.. it has trouble unpacking functions.

akdetrick (changes requested):
Few more small things, but looks like a major improvement over the previous interface for Popover

akdetrick (approved):

PR #532, opened 2020-12-24
UIP-642: MenuLink

reywright:
## Description
Expose MenuLink Component to be used alongside MenuItem inside FDS Menu

## Screenshots
Screen Shot 2020-12-24 at 12 43 28 PM

## Checklist
- [x] All unit tests pass


src/components/MenuLink/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+  children: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.node), PropTypes.node])
+    .isRequired,
+
+  /** Styles menu item as disabled and disables selection events */
+  isDisabled: PropTypes.bool,
+
+  href: PropTypes.string,
+
+  /** Useful for passing react router Link */
+  as: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
reywright, line 33: soo..... `MenuLink` accepts rest props. `href` isn't explicitly defined and is merely passed as a rest prop, HOWEVER. I feel like it's important to specify it explicitly here else people are going to be confused. However, the `as` prop IS explicitly defined, and it'll be obvious to whoever uses that, that a `to` prop can obviously be passed without having to spell it out, and without incurring upfront confusion ("question: what's the difference between href and to"). Not married to this though and happy to make changes.

src/components/MenuLink/index.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { MenuLink as ReachMenuLink } from '@reach/menu-button';
+
+const noop = () => {};
+
+const MenuLink = ({ onSelect, children, isDisabled, href, as }) => (
```
akdetrick, line 7: should `as` have a default value in our wrapper?

src/components/MenuLink/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+  onSelect: PropTypes.func.isRequired,
+
+  /** React children - any JSX can be used in a menu item */
+  children: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.node), PropTypes.node])
+    .isRequired,
+
+  /** Styles menu item as disabled and disables selection events */
+  isDisabled: PropTypes.bool,
+
+  href: PropTypes.string,
```
akdetrick, line 30:
```suggestion
  /** When the `as` prop is set to use an `a` element, use this to set the `href` attribute */ 
  href: PropTypes.string,
  
  /** When the `as` prop is set to use a `Link` element, use this to set the `to` prop of the `Link` */
  to: PropTypes.string,
```

src/components/MenuLink/index.jsx
akdetrick: What do you think of my suggestion above (documenting both `href` and `to` explicitly)?

akdetrick (changes requested):
Just a few things to clear up around `as`, `to`, and `href`, otherwise looks good

src/components/MenuLink/index.jsx
reywright: as is only really needed to use `Link`. Leaving it undefined, reach will render it as an anchor, so I didn't find it necessary.

src/components/MenuLink/index.jsx
reywright: This would technically be something like 

`/** When the `as` prop is set to use an `a` element or is undefined, use this to set the `href` attribute */`

And at that point I wonder if that makes it more confusing. I feel like if someone wanted to use MenuLink to render an anchor, if they see an href, they would assume adding the href outright works (which is true).

src/components/MenuLink/index.jsx
reywright: I'm cool with that, I'll make that change. I may need to circle back to see if I did what you were expecting though.

akdetrick (approved):
👏

PR #533, opened 2020-12-24
UIP-812: menuitem isActive

reywright:
## Description
Add isActive state to MenuItem.

I dropped this on the ticket but it bears repeating: 

ReachMenuItem nor the ReachMenu component support an active state. Adding an active state here (in the facade) means that design knows it exists, and they’ll want 100% of menu’s to be using active states. To use active states, developers will need to add isActive to all MenuItems and manage the conditionals themselves to ensure the correct item is marked by them. That seems obvious but I guarantee it won't be and it will be extra work by developers to add this. So the question is, if Reach doesn't support this, is this even a feature we want to support, that will moreso create more work for us with minimal payoff. Something carefully to think about before we try to merge this.

## Screenshots
Screen Shot 2020-12-24 at 1 28 44 PM

## Checklist
- [x] All unit tests pass


akdetrick:
@rey-wright I agree with your concerns. The `Menu` component should be reserved for menus. I think this feature was requested because we were treating menus like a `select` in some places.

This is not something we want to encourage. I'd recommend we close this for now and address this as part of our Q1 goals

reywright:
🎉 😅

PR #571, opened 2021-01-07
156/rocket 1085 popovers and dialogs from portals

sfrieson-cbi:
## Description
In trying to update the cbi-tables `HeaderCell` dropdown to use the FDS `Popover` component instead of the local `Modal` component, a few problems came up. This MR adds solutions to resolve those issues.

## Issues
Here are the two main issues that I ran into and how this MR supports their solution.

### `Dialog`/`Prompt`

image

#### Problem

There are UXs in Collections that support opening an alert `Prompt` from a popover. You will notice in the screenshot above, behind the `Prompt` there is a `Popover` with several `x` icons which allows the user to delete steps. Since the user has the opportunity to make several different actions in this `Popover`, it should not close after any atomic action, but since this `Prompt` is in a portal the `Popover` hears clicks in the `Dialog` as outside clicks and unmounts itself and subsequently the `Prompt`. This also prevents the confirmation in the `Dialog` from succeeding.

**Solution**
A `disablePortal` prop has been added to the `Dialog` and `Prompt` components to support this use case.

___

### `Popovers` inside `Popovers` with `overflow: scroll;`

image

#### Problem
With funnel/step columns, you can edit the steps (as seen previously) inside of the `Popover`. The edit UI is within the same `Popover` and slides in from the right side of the `Popover`, so it needs `overflow: scroll;` to hide the offscreen portions. The reasons behind that choice are slightly unrelated but can be expanded upon if desired. The `ColorPicker` in the edit UI is also in a `Popover`. It cannot be in a portal for the same reasons as the issue above with the `Prompt`. When the portal is disabled, the `Popover` becomes subject to the `overflow: scroll;`.

#### Solution
The usage of `position: fixed;` breaks out of the `overflow: scroll;` bounds, though I'm not entirely sure why. An `isFixed` prop has been added to the `Popper` to allow this stying.

___

## Checklist
- [x] All unit tests pass


src/components/Popover/index.jsx
```diff
@@ -14,8 +14,15 @@ const Popover = ({
...
   appendTo = () => document.body,
+  isFixed = false,
 }) => {
   const hideOnEsc = useMemo(() => hideOnEscFunc(onUserDismiss), [onUserDismiss]);
+  const popperOptions = useMemo(
+    () => ({
+      strategy: isFixed ? 'fixed' : 'absolute',
+    }),
+    [isFixed]
+  );
```
reywright, line 25: 👍

reywright (approved):
nice! thanks.

reywright:
`position: fixed` essentially positions on the viewport itself, which is why it's able to break out of overflow hidden. Thankfully, event bubbling still works the way you'd expect it to.

PR #574, opened 2021-01-13
[V10] UIP-1015: Toaster API changes

reywright:
## Description
- toastProps changed to toastInstance
- onDismiss moved inside toastInstance
- dismissDelay removed 
- documentation changed to be much more clearer and robust to show toaster functionality (and to remove all references to private Toast component)

## Screenshots
Screen Shot 2021-01-13 at 10 45 28 AM

## Checklist
- [ ] All unit tests pass

### DIFF
--- a/src/components/Toaster/index.jsx
+++ b/src/components/Toaster/index.jsx
@@ -5,8 +5,9 @@ import { CSSTransition, TransitionGroup } from 'react-transition-group';
 import customPropTypes from 'components/util/customPropTypes';
 import Toast, { TYPES } from './Toast';
 
+const dismissDelay = 4000;
+
 export const Toaster = ({
-  dismissDelay = 4000,
   isAutoDismiss = true,
   isOpen = false,
   toastInstance = {},
@@ -84,8 +85,6 @@ Toaster.propTypes = {
   isOpen: PropTypes.bool,
   /** Should this toast auto-dismiss itself? */
   isAutoDismiss: PropTypes.bool,
-  /** Time in ms to auto-dismiss toast */
-  dismissDelay: PropTypes.number,
   /** Controls the display for your specific toast. */
   toastInstance: PropTypes.shape({
     /** JSX Content of Toast */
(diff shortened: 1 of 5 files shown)


src/components/Toaster/index.jsx
```diff
@@ -2,33 +2,42 @@ import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
 import ReactDOM from 'react-dom';
 import { CSSTransition, TransitionGroup } from 'react-transition-group';
-import Toast from './Toast';
+import customPropTypes from 'components/util/customPropTypes';
+import Toast, { TYPES } from './Toast';
+
+const dismissDelay = 4000;
```
reywright, line 8: This is now just a non configurable constant. Nothing else has to change.

sfrieson-cbi: Just an FYI, do you know that [work was just done a couple of weeks](https://cbinsights.slack.com/archives/CD4CK5M7E/p1609800619489400) to make this work correctly with a configurable?

reywright: Yessir! This PR stems directly from those points (paradoxically):

- Apparently the delay was broken for around 9 months and no one noticed, which leads us to believe that it actually isn't that important.
- From what Adam said, there's about 8 instances in the platform that are using customized delays, and a lot of them are close enough to 4 seconds that we would want them to be 4 seconds.
- Creating different delays is sort of arbitrary and potentially would lead to bikeshedding. Moreover, we're making assumptions with longer text as to how fast the user can read (especially if they have a hard time reading).

Given all these bullet points, it made more sense to us to have a single delay, and if people need the Toaster to show for a longer time, then to merely turn isAutoDismiss off.

cc @akdetrick

reywright: I also worked closely with @kchen-cbi to figure out the correct api so he's in the loop and up to date with these changes / decisions as well.

sfrieson-cbi: >  Moreover, we're making assumptions with longer text as to how fast the user can read (especially if they have a hard time reading).

Future feature request: To that point, it might be worth thinking about supporting pausing the auto dismiss until the user's cursor is no longer over the toast. This is the UX macOS notifications have, which I personally use often. It probably allows the notification to be compliant with [WCAG 2.2.1 (level A)](https://webaim.org/standards/wcag/checklist#sc2.2.1).

akdetrick (approved):
🍞

PR #599, opened 2021-02-10
Prettier lint staged

reywright:
## Description
- Updated lint-staged to latest version (along with configuration)
- Updated prettier to latest version
- Prettier configuration changed so that now it's being applied to ALL files (with the consideration that it will ignore anything in the `.prettierignore` file, and also any file that prettier itself does not want to run on )
- stylelint changed so that it does not make formatting fixes, reverting changes made by prettierr (this was mostly done using stylelint-recommended-config
- Finally, ran prettier on the entire codebase and committed that apart of this as well.

I would love if people could pull this down, make commits changing javascript / css, and see if this works for them.

## Checklist
- [ ] All unit tests pass


package.json
```diff
@@ -139,25 +140,21 @@
...
-      ]
-    },
-    "ignore": [
-      "src/**/*.test.{js,jsx}",
-      "src/**/assets/*.css"
+    "src/**/*": [
+      "prettier --write"
+    ],
+    "src/**/*.{js,jsx}": [
+      "eslint --no-ignore"
```
sfrieson-cbi, line 147: I think you want a `--fix` on this one.

reywright: you know... I was under some assumption that our eslint would override some of the prettier settings, but I didn't realize that our eslintrc.js is actually extending the eslint prettier plugin. Good catch, I'll fix this.

akdetrick (approved):

PR #618, opened 2021-02-22
UIP-1025: LazyTippy

reywright:
## Description

Tippy renders content immediately even if it's not triggered or visible in the DOM. We absolutely cannot use Tippy if that is the case.

The maintainer thankfully pointed me to an implementation that allows us to get it to render lazily:

https://github.com/atomiks/tippyjs-react/issues/288
https://gist.github.com/atomiks/520f4b0c7b537202a23a3059d4eec908

I have added this component and consumed it in Popover and Tooltip. Couple of notes:

- the lazy implementation breaks `moveTransition` so I have nuked this for now (it's a really nice to have but not crucial)
- this required changing the tests to be async. I've went through and commented out the relevant parts.

## Screenshots

The one on the right is LazyTippy with moveTransition, which is why I removed support for it:

## Checklist

- [x] All unit tests pass


package.json
```diff
@@ -152,7 +152,7 @@
   },
   "version": "4.0.0",
   "dependencies": {
-    "@popperjs/core": "^2.5.4",
+    "@popperjs/core": "^2.7.2",
```
reywright, line 155: There's no real impetus behind this upgrade other than that it needs to be done sometime.

src/components/Popover/LazyTippy.js
```diff
@@ -0,0 +1,31 @@
+import React, { useState } from 'react';
+import Tippy from '@tippyjs/react';
+
+// Will only render the `content` or `render` elements if the tippy is mounted to the DOM.
+// Replace <Tippy /> with <LazyTippy /> component and it should work the same.
+
+/* eslint react/prop-types: 0 */
```
reywright, line 7: I really don't fee like this component should have prop types as that could create issues later on if Tippy updates it's proptypes, these could get out of sync.

src/components/Popover/LazyTippy.js
```diff
@@ -0,0 +1,31 @@
...
+  if (props.render) {
+    computedProps.render = (...args) => (mounted ? props.render(...args) : '');
+  } else {
+    computedProps.content = mounted ? props.content : '';
+  }
+
+  return <Tippy {...computedProps} />;
+}
+
+export default LazyTippy;
```
reywright, line 31: This entire component is stolen directly from the gist here: https://gist.github.com/atomiks/520f4b0c7b537202a23a3059d4eec908

src/components/Popover/index.test.js
```diff
@@ -1,27 +1,27 @@
...
-  it('tests that trigger and content text gets rendered', () => {
+  it('tests that trigger and content text gets rendered', async () => {
     render(
       <Popover isOpen triggerElement={<button>trigger</button>}>
         <p>popover content</p>
       </Popover>
     );
-    expect(screen.getByText('popover content')).toBeTruthy();
     expect(screen.getByText('trigger')).toBeTruthy();
+    await waitFor(() => expect(screen.getByText('popover content')).toBeTruthy());
```
reywright, line 25: `async`, `await`, and `waitFor` ALL must be used here in order to maintain the integrity of the test now that `LazyTippy` is being used. If they are not, this will trigger a bunch of `act` errors and "popover content" also won't be present in the DOM tree.

src/components/Tooltip/index.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import FDS from 'lib/dictionary/js/styleConstants';
-import Tippy from '@tippyjs/react';
+import LazyTippy from 'components/Popover/LazyTippy';
```
reywright, line 4: We should probably make some `private` folder to put common abstractions between components, while at the same time communicating that people should absolutely not import them, but that probably requires discussion and strategizing that may blow out the scope here. I just decided it would be easier for LazyTippy to live in `Popover` and we can come back and do an organization ticket for the rest.

src/components/Tooltip/index.test.js
```diff
@@ -1,13 +1,15 @@
...
 describe('Tooltip component', () => {
-  it('matches snapshot (uses the right classes)', () => {
-    const wrapper = mount(<Tooltip trigger={<Trigger />} message="lol" />);
-    expect(wrapper).toMatchSnapshot();
+  it('tests that message gets rendered on hover', async () => {
+    render(<Tooltip trigger={<button>trigger</button>} message="lol" />);
+    userEvent.hover(screen.getByText('trigger'));
+    await waitFor(() => {
+      expect(screen.getByText('lol')).toBeTruthy();
+    });
```
reywright, line 13: The importance of creating tests like this cannot be understated. One thing to note, is that in the snapshot test, the message of "lol" should NOT be rendering as it hasn't been hovered. Moving this to a testing-library test not only makes this more painfully obvious, it makes it easy to add a hover event to properly trigger it. This is unlikely to be a false positive / false negative moving forward.

src/components/Popover/LazyTippy.js
akdetrick: There are still advantages for this component to explicitly declare its dependencies and document them. We can do it in a follow up though

src/components/Tooltip/index.test.js
akdetrick: The old test and the new test are testing different things (class name application vs if it gets rendered at all) - but point taken, yours is a more useful test

akdetrick (approved):

sfrieson-cbi:
Thanks for researching this and fixing Rey!

PR #631, opened 2021-03-01
Move all class documentation into single mdx file

reywright:
## Description

I would like to label 2 things as out of scope as it pertains to this PR:

- The naming of each of the sections (those names already exist)
- The usefulness of any of these sections (as these already exist)

To me, the thing that's most relevant to discuss here prior to considering to merge are

1. The DX of merging all stories documents into one document
2. How the UX feels in terms of navigating around.
3. Any more technical discussions around grouping (which Storybook is limited on)

## Screenshots

Screen Shot 2021-03-01 at 10 35 35 AM

## Checklist

- [ ] All unit tests pass


reywright:
To be clear, we can absolutely rename / delete sections but this would be super easy to do later. I think the cleanest thing is to punt on that discussion now and focus solely on the mechanics of grouping these together, and then polish up the naming at another time.

akdetrick (approved):
👏

PR #636, opened 2021-03-04
UIP-1032: remove classnames peer dependency

reywright:
## Description

We no longer depend on classnames.

## Screenshots

N/A


akdetrick (approved):
good catch!

PR #637, opened 2021-03-05
[v12] UIP-1033: new modal engine (Reach UI)

reywright:
## Description

Move Dialog over to Reach UI + framer motion.

I'm marking this as **Don't Code Review** till we have our v10/v11 retrospective @akdetrick 

Benefits

- Better animations
- Less code to maintain
- Better thoughtfulness around accessibility

This IS a breaking change:
- Reach UI does not support disabling of the portal (non issue as we don't use this in the site)
- Reach UI does not support adding a role, which it claims introduces unnecessary issues (also non issue as we don't use this in the site)
- Removed ability to pass width and height in number values (interpolated as pixel values. long since deprecated and removed from documentation)

Other changes

- [Performance] Stop creation of new components in render
- [Performance / Functionality] Eschew resize listener (which doesn't actually do what we want fully) in favor of `ResizeObserver`. Will create a presentation for the team to discuss the massive benefits of this approach.
- [Organization] Split up components, while avoiding prop drilling by using React Context
- [Organization] Split up the larger Dialog component into smaller digestable components and files (`DialogHeader`, `DialogContent`, `DialogFooter`, `DialogAnimation`, and `DialogContext`)

## Screenshots

N/A

## Checklist

- [x] All unit tests pass


package.json
```diff
@@ -153,11 +153,13 @@
   "version": "4.0.0",
   "dependencies": {
     "@popperjs/core": "^2.7.2",
+    "@reach/dialog": "^0.13.2",
     "@reach/menu-button": "^0.12.1",
     "classcat": "^4.1.0",
-    "focus-trap-react": "^8.3.2",
+    "framer-motion": "^3.10.0",
```
reywright, line 159: Couple of things:

* The api for css transition is pretty outdated and not so great.
*  Ryan Florence (creator of Reach UI) refused to support it for their dialog. So we can't use it even if we wanted to.
* Couple of libraries for animation available. react-spring is nice but has memory leak issues, and many months ago I tried this same PR with react-spring and animations were not completing in staging.
* framer-motion is probably the most accessible, up to date, maintained, and powerful animation library, so this is what I'm opting to use here.

src/components/Dialog/components/DialogHeader.jsx
```diff
@@ -0,0 +1,31 @@
...
+import DialogContext from '../DialogContext';
+
+const DialogHeader = () => {
+  const { onDismiss, title } = useContext(DialogContext);
+  return (
+    <>
+      {(title || onDismiss) && (
+        <div className="bgColor--white border--bottom">
+          <div className="dialog-header padding--left padding--top padding--bottom">
+            <div className="type--head4">{title}</div>
```
reywright, line 13: TODO: come back and ensure there's an empty string here when no title is passed.

src/components/Dialog/index.test.js
```diff
@@ -23,6 +20,8 @@ describe('Dialog', () => {
       />
     );
 
+    expect(screen.getByLabelText('Close')).toHaveFocus();
```
reywright, line 23: Going to start doing a lot more focus management & focus testing in these tests.

src/components/Dialog/index.test.js
```diff
@@ -31,9 +30,9 @@ describe('Dialog', () => {
 
     /* Tests that onDismiss gets fired correctly */
     /* ========================================= */
-    fireEvent.click(screen.getByLabelText('Close'));
+    userEvent.click(screen.getByLabelText('Close'));
     // Pressing esc anywhere in dialog should trigger a close
-    fireEvent.keyDown(screen.getByText('content'), { key: 'Escape', code: 27 });
+    userEvent.type(screen.getByText('content'), '{esc}');
```
reywright, line 35: it may be the case that `fireEvent` should never be used alongside `userEvent` as it encompasses all the methods of the other.

src/components/Dialog/index.test.js
```diff
@@ -44,7 +43,5 @@ describe('Dialog', () => {
     expect(window.getComputedStyle(screen.getByRole('dialog'))['max-width']).toBe(
       '1000px'
     );
-    // Scrolling should be locked when dialog is open
-    expect(window.getComputedStyle(document.documentElement).overflow).toBe('hidden');
```
reywright, line 48: To my surprise (and actually when I think about, this is smart) Reach UI Dialog does not apply overflow hidden to the body, but merely locks the scrolling on body. The most powerful part of this is not having to deal with a shifting page as a scrollbar enters and leaves (our current implementation) or, what most implementations do, try to sniff the width of the scrollbar and add commensurate right padding to the screen so that things don't shift around.

src/components/Dialog/useResizeObserver.js
```diff
@@ -0,0 +1,25 @@
...
+
+  useLayoutEffect(() => {
+    observe();
+    return () => disconnect();
+  }, [disconnect, observe]);
+
+  return [setNode, observerEntry];
+};
+
+export default useResizeObserver;
```
reywright, line 25: Technically I want to just download a hook for this, but all of them didn't work. That being said, it's likely they didn't work due to a false negative I discovered (and fixed) later working on this component. I'll need to revisit it but not incredibly pressing.

PR #641, opened 2021-03-12
UIP-1036: button blue outlined

reywright:
## Description

Add button blue outline

## Screenshots

Screen Shot 2021-03-12 at 12 31 03 PM

## Checklist

- [x] All unit tests pass


src/components/style/button.css
```diff
@@ -227,6 +249,24 @@
 .fdsButton--outlined.fdsButton--isDestructive.fdsButton--isActive:focus {
   background-color: #fad6d8;
 }
+
+.fdsButton--outlined--blue.fdsButton--isDestructive.fdsButton--isActive {
```
akdetrick, line 253: do we want think pinkish color for active state of `blue-outlined`?

akdetrick: looks like this gets reset later in the cascade - do the outlined blue variant classes need to be in these selectors?

akdetrick (commented):
Just a CSS question before we move forward

src/components/style/button.css
reywright: This is for the `isDestructive` modifier, which is for all intents and purposes a separate theme / theme modifier.

Screen Shot 2021-03-12 at 5 24 34 PM

We may need to rethink this button api holistically though as we discussed, but there doesn't seem to be an ideal choice, just what tradeoffs are we happiest with 😅

akdetrick: I forgot it was a separate prop! Works for me

akdetrick (approved):

akdetrick (changes requested):
almost missed this - needs a version bump + changelog entry

PR #651, opened 2021-04-03
UIP-1048: related components

reywright:
## Description

* Add plumbing for related components
* Added a few related sections. Further related sections can be added in a followup PR.

## Screenshots

Screen Shot 2021-04-02 at 7 14 24 PM

## Checklist

N/A

### DIFF
--- a/src/util/storybook-docs/Related.jsx
+++ b/src/util/storybook-docs/Related.jsx
@@ -1,0 +1,27 @@
+import React, { useContext } from 'react';
+import { DocsContext } from '@storybook/addon-docs/blocks';
+
+export const Related = () => {
+  const context = useContext(DocsContext);
+
+  const related = context.parameters.docs.fds?.related;
+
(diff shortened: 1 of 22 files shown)


src/util/storybook-docs/Related.jsx
```diff
@@ -0,0 +1,28 @@
...
+export const Related = () => {
+  const context = useContext(DocsContext);
+
+  const related = context.parameters.docs.fds?.related;
+
+  if (!related) return null;
+
+  return (
+    <div className="related">
+      Are you looking for:
```
akdetrick, line 14: I actually like "Related" copy here better - it's a bit more neutral. To me "Are you looking for" sounds like an indication of a mistake when there might not be one

akdetrick (approved):
Nice use of CSF to make this work! I have one copy suggestion, but otherwise good to go

PR #653, opened 2021-04-05
UIP-1063: Fix Link prop types to use elementType instead of func

kchen-cbi:
## Description
[Reach UI's MenuLink requires any Link component passed to it to use the forwardRef API](https://reach.tech/menu-button/#menulink), but our MenuLink enforces PropTypes.func for the Link prop, which will break for any Link using forwardRef (which returns a React component object). [Since `elementType` is available to us as of prop-types v15.7.0](https://github.com/facebook/prop-types/blob/v15.7.2/CHANGELOG.md#1570), we should opt to use it instead.

## Checklist

- [x] All unit tests pass


reywright (approved):
ooo fancy

PR #654, opened 2021-04-13
UIP-1065: moved all references to jest-dom extend-expect to one place

reywright:
## Description

add jest-dom extend to the setup so that we don't have to keep doing it in each test


akdetrick (approved):
🎉

PR #658, opened 2021-04-23
Loosens unnecessarily strict prop type

sfrieson-cbi:
## Description

There was [a recent update to the `Dialog` component](https://github.com/cbinsights/form-design-system/pull/657) which allowed the `content` property to be conditionally rendered, but the prop type for it remained as required. This results in prop type warnings for clients as well as the `Prompt` component which can supply `content={null}`.

This MR removes the required constraint.

## Checklist

- [x] All unit tests pass


akdetrick (approved):

akdetrick:
🙏 thank you for the fix

PR #662, opened 2021-05-10
UIP-1073: add package that throws errors on prop type warnings when running tests

reywright:
## Description

We recently had a situation where a proptype was incorrect in FDS: https://github.com/cbinsights/form-design-system/pull/658.

This proptype was throwing an error in `cbi-site`. It was throwing in `form-design-system` tests but it was an optional warning. Given that `cbi-site` throws errors on proptypes in test, we should throw errors on proptypes in tests.

## Screenshots

N/A

## Checklist

N/A


sfrieson-cbi (approved):
Thanks for adding this!

PR #675, opened 2021-06-08
Introduction of Typescript to FDS (plus conversion of Button)

reywright:
## Description

* Adds Typescript to codebase
  * Babel typescript plugin being used
  * Change to Typescript eslint parser
  * Adds CI check
* Converts Button folder in components completely over to Typescript

Pending testing this in staging, this seems ready to go.

One thing to note: full typescript bundling (sans babel) will be revisited when the initiative to move to Rollup happens.

## Screenshots

N/A

## Checklist

- [x] All unit tests pass


src/components/Button/index.test.tsx
```diff
@@ -6,12 +6,7 @@ import Button from '.';
 describe('Button component', () => {
   it('tests default rendering', () => {
     render(<Button label="Button" />);
-    const button = screen.getByRole('button', { name: 'Button' });
-    expect(button).toBeTruthy();
-    expect(button).toHaveClass('fdsButton--m');
-    expect(button).toHaveClass('fdsButton--blue');
-    expect(button).not.toHaveClass('fdsButton--isDestructive');
-    expect(button).not.toHaveClass('fdsButton--isFullWidth');
```
reywright, line 14: hmmmm why did I remove this....

src/util/storybook-docs/StoryLayout.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import FDS from 'lib/dictionary/js/styleConstants';
 
-export const StoryWrapper = ({ children, className, ...rest }) => (
+export const StoryWrapper = ({ children, className = '', ...rest }) => (
```
reywright, line 5: important to note: importing this non Typescript file into a Typescript file caused an error to be throw, because I wasn't passing a className. It's interesting that Typescript interprets a destructure as *required* props automatically.

tsconfig.json
```diff
@@ -0,0 +1,31 @@
+{
+  "compilerOptions": {
+    "baseUrl": "./src",
+    "paths": {
+      "lib/*": ["../lib/*"],
```
reywright, line 5: This needs to go up one level as the paths here are based on the baseUrl (this was just a random guess I made, but it turned out to fix the issue)

.gitlab-ci.yml
```diff
@@ -20,6 +20,17 @@ include:
...
+  retry: 2
+  stage: test
+  script:
+    - yarn type-check
+  tags:
+    - shell
+    - dev
+  except:
+    - tags
+
```
reywright, line 33: I'm not attached to this name but we should get the name here correct (type-check, typescript, typescript-check, etc), and I can file a devops ticket and give them this yml file partial to put in gitlab includes

.eslintrc.js
```diff
@@ -1,32 +1,40 @@
...
+    '@typescript-eslint/no-empty-function': 0, // We like using empty noop (for now)
+    '@typescript-eslint/explicit-module-boundary-types': 'off', // It's necessary to turn this off for non Typescript files
+    /* =========================================
+
+      We use "no-restricted-syntax" is used to
+      limit what gets rolled out in each stage.
+      https://cbinsights.atlassian.net/wiki/spaces/TD/pages/2379907163/Rollout+plan
+
+      ========================================= */
+    'no-restricted-syntax': [
```
reywright, line 24: It seems obvious now, but just make sure this is inside `rules` and that there isn't 2 sets of `rules` or issues will occur.

.eslintrc.js
```diff
@@ -1,32 +1,40 @@
...
-  extends: ['airbnb-base', 'plugin:react/recommended', 'prettier'],
-  plugins: ['react', 'jest'],
+  parser: '@typescript-eslint/parser',
+  extends: [
+    'eslint:recommended',
+    'plugin:@typescript-eslint/recommended',
+    'plugin:react/recommended',
+    'plugin:jest/recommended',
+    'prettier',
+  ],
```
reywright, line 10: moving more to recommended configs, I was able to get rid of most of our custom ones.

.storybook/main.js
```diff
@@ -1,9 +1,9 @@
 const path = require('path');
 
 module.exports = {
-  stories: ['../src/**/*.stories.@(js|mdx)', '../guidelines/**/*.stories.@(js|mdx)'],
-  addons: ['@storybook/addon-postcss', '@storybook/addon-essentials', '@storybook/addon-a11y'],
```
reywright, line 5: this shouldn't have been removed....

.eslintrc.js
```diff
@@ -1,32 +1,40 @@
...
+    ],
   },
   overrides: [
     {
-      files: ['*.test.js'],
+      // enable the rule specifically for TypeScript files
+      files: ['*.ts', '*.tsx'],
       rules: {
-        'import/no-unresolved': false,
+        '@typescript-eslint/explicit-module-boundary-types': ['error'], // We want to start throwing again on Typescript files
```
sfrieson-cbi, line 37: 💯

.gitlab-ci.yml
sfrieson-cbi: I'm onboard with a more generic name (no "typescript"). This step is about static type checks. I like it.

.gitlab-ci.yml
```diff
@@ -20,6 +20,17 @@ include:
     ref: latest
     file: 'front-end-publish-npm.yml'
 
+type-check:
+  retry: 2
```
sfrieson-cbi, line 24: I don't think we need retries do we?  This isn't a huge deal, but type checking shouldn't be flakey. Did you come across an actual need for this?

.storybook/main.js
```diff
@@ -1,9 +1,9 @@
...
-  webpackFinal: async (config) => {
+  stories: ['../src/**/*.stories.@(js|ts|mdx|tsx)', '../guidelines/**/*.stories.@(js|ts|mdx|tsx)'],
```
sfrieson-cbi, line 4: This is reminding me that we never had a conversation about `.tsx`. [Some people don't like it](https://github.com/airbnb/javascript/pull/985#issuecomment-239145468). We currently have mixed implementations between our repos and I wonder if now is a good time to try to standardize it and get it into a style guide.

package.json
```diff
@@ -27,17 +27,18 @@
...
     "update-snapshots": "jest -u --color",
-    "lint": "eslint ./src/ --ext .js,.jsx",
+    "lint": "eslint ./src/ --ext .js,.jsx,.ts,.tsx",
     "precommit": "lint-staged",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
     "stats:css": "node ./scripts/node/util/getClassStats.js",
-    "stats:imports": "node ./scripts/node/util/getImportStats.js"
+    "stats:imports": "node ./scripts/node/util/getImportStats.js",
+    "type-check": "tsc --jsx react --allowSyntheticDefaultImports --noEmit"
```
sfrieson-cbi, line 41: I feel like this command should be able to be simplified to `"tsc"`. It should be able to grab all of the flag values from the `tsconfig.json`.

src/components/Button/index.stories.tsx
```diff
@@ -105,13 +107,14 @@ export const isActive = () => (
...
             <>
-              {['ghost', 'outlined'].map((theme, idx) => (
+              {INVERTED_THEMES.map((theme, idx) => (
                 <StoryItem key={idx}>
                   <Button
                     theme={theme}
                     isActive={!value[idx]}
                     label={theme}
                     onClick={() => setValue({ ...value, [idx]: !value[idx] })}
+                    data-test="hey"
```
sfrieson-cbi, line 117: Do you still want this here?

src/components/Button/index.stories.tsx
```diff
@@ -60,7 +62,7 @@ Inverted.parameters = {
   },
 };
 
-export const isActive = () => (
+export const isActive = (): JSX.Element => (
```
sfrieson-cbi, line 65: I've been typing the component as opposed to the return type:
```suggestion
export const isActive: React.FC = (): JSX.Element => (
```
`React.FC` is an alias for `React.FunctionComponent`. This could be more flexible because you can add props later if you need to, i.e. `React.FC<ComponentPropInterface>`.
The return type is also _slightly_ different. It returns a `ReactElement | null`.  `ReactElement`'s definition looks like this:
```typescript
interface ReactElement<P = any, T extends string | JSXElementConstructor<any> = string | JSXElementConstructor<any>> {
    type: T;
    props: P;
    key: Key | null;
}
```

I only bring all this up because it might be good to be consistent about this sort of thing.

src/components/Button/index.stories.tsx
```diff
@@ -214,7 +220,7 @@ Loading.parameters = {
   },
 };
 
-export const Breakout = (args) => (
+export const Breakout = (args: ButtonProps): JSX.Element => (
```
sfrieson-cbi, line 223: Hmm... I see here that this might go against my comment above. I wonder if Storybook has a type they offer for this.

src/components/Button/index.tsx
```diff
@@ -1,13 +1,76 @@
...
 import baseElement from 'util/baseElement';
 import { CaretDownIcon } from 'lib/icons/react';
 
-export const SIZES = ['s', 'm'];
-export const THEMES = ['blue', 'orange', 'outlined', 'outlined--blue', 'ghost'];
+export const SIZES = ['s', 'm'] as const;
 
-const Button = forwardRef(
+export const THEMES = ['blue', 'orange', 'outlined', 'outlined--blue', 'ghost'] as const;
+export type THEME = typeof THEMES[number];
```
sfrieson-cbi, line 10: Stylistically, I think this should be title-cased:

```suggestion
export type Theme = typeof THEMES[number];
```

src/components/ButtonGroup/index.jsx
```diff
@@ -5,33 +5,17 @@ import GroupButton from './GroupButton';
...
-        {
-          value,
-          content,
-          key,
-          Icon,
-          onClick,
-          label,
-          // `isFirstButton` and `isLastButton` are private and can't being passed in.
-          isFirstButton,
-          isLastButton,
```
sfrieson-cbi, line 19: These two props aren't getting carried over?

.gitlab-ci.yml
reywright: I can remove it: I just copied this from another step I found.

src/components/Button/index.stories.tsx
reywright: Any thoughts on what this resource says? https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/ (it has reasoning why it recommends against React.FC down below)

src/components/ButtonGroup/index.jsx
reywright: When I updated eslint, it complained that these values were not being used, which I don't think they are.

.storybook/main.js
reywright: Discussed offline. We want to move to using strictly js & ts.

src/components/Button/index.stories.tsx
reywright: fixed 🙇

src/components/Button/index.stories.tsx
sfrieson: I'm on board with the reasons brought up in that article and in secondary links. My two main takeaways:
- it's more explicit what you want the return to me. Some people might see it as a chore, but it'll hopefully help them understand better.
- The `React.FunctionComponent` by default allows your component to accept `children` even if you don't provide them in the props. That's just misleading.

I think the main reason I've always used `React.FC` is because when I first started using TS + React I found it hard to figure out which return type to use: `React.Element`, `React.ElementType`, `React.ReactNode`. They all sound vaguely appropriate if you're just searching through autocomplete for what to use, but I'd never guess `JSX.Element`, so when I found that `React.FC` just took care of it for me, I started using that.

.storybook/main.js
sfrieson: Followed up further offline, there are some limitations to the ts compiler at the moment that force us to use `.tsx`.

.storybook/main.js
reywright: ohhh it's implicit and will be deprecated in v7. Adding back now.

sfrieson (approved):
Thanks for the updates! Super exciting about getting TypeScript in here. Have we tested the beta version in cbi-site yet?

reywright:
@sfrieson thank you kindly for the code review! I put a beta on staging on July 14th, and @jackmac92 pinged me on July 18th asking me if he can remove the beta from staging, so there was a beta that stayed on staging for 4 days, and I didn't hear a peep otherwise 🙂 

I'm thinking once this releases, I'll throw it again on staging (or dev) just to double double check as there's no feature here, hence no pressure to immediately add it to `cbi-site`.

src/components/Button/index.stories.tsx
reywright: My perception is that the story itself doesn't have much to do with storybook itself... it's just a component for the most part, but yeah let me know if you find anything.

PR #685, opened 2021-09-28
add headerFooter prop to dialog component

clew-cbi:
## Description

Add a headerFooter prop to the Dialog component
## Screenshots

Screen Shot 2021-09-28 at 2 03 26 PM

## Checklist

- [x] All unit tests pass


src/components/Dialog/index.jsx
```diff
@@ -193,6 +197,9 @@ Dialog.propTypes = {
 
   /** Disables rendering the dialog in a portal, and renders it locally instead. */
   disablePortal: PropTypes.bool,
+
+  /** Add content below the title */
+  headerFooter: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
```
sfrieson, line 202: I'm finding this prop name to be confusing. What do you think of `subTitle` or `headerBody`.

I think I like `subTitle` because the location it goes is clear, and the props explain that you can supply more than strings.

___

You can simplify the PropTypes to `PropTypes.node`.  `node` will take care of `PropTypes.string` and `PropTypes.element`.

sfrieson (commented):
Hey Chris, Thanks for putting this up. I've left a couple comments and suggestions inline.

src/components/Dialog/index.jsx
clew-cbi: `subTitle` sounds good to me. I'll update the PropTypes too. Thanks for checking the MR out.

sfrieson (approved):
Thanks for those updates!

PR #686, opened 2021-09-29
Move import order for form modifers above spacing helpers.

jjniczCB:
## Description

Updating the import order for the `Form field modifiers` to be above `spacing helpers`.
This fixes an issue where you could not combine some spacing utility classes with `resetButton` or `resetInput`.

Currently, if you wanted to do something like className="resetButton padding--x", the .resetButton class overrides any other padding helper classes you add.

## Checklist

- [x] All unit tests pass


sfrieson:
Thanks for waiting on this one @jjniczCB. Since we don't have the pipeline publishing beta versions a the moment it's a little hard to be super confident in these changes. I did however go in and compare the classes you moved to the ones that we reordered. It seems to me that you have not introduced any cascade issues (though I did uncover some existing specificity bugs).

I'm good to go with this once you pull in `master` and rebuild to resolve the merge conflict.

PR #697, opened 2021-11-03
Migrate Avatar to Typescript

mwahlmann-cbi:
## Description

Migrated the Avatar component, tests and storybook to Typescript. I've also created an Icon and type that can be used in Prop interfaces to ensure an Icon is passed.

## Checklist

- [X] All unit tests pass

### DIFF
--- a/src/components/Avatar/index.test.tsx
+++ b/src/components/Avatar/index.test.tsx
@@ -37,38 +37,37 @@ describe('grabInitials', () => {
 describe('Avatar component', () => {
   it('renders name', () => {
     render(<Avatar name="Joey Tribbiani" />);
-    expect(screen.getByTitle('Joey Tribbiani', { name: 'JT' })).toBeTruthy();
+    expect(screen.getByTitle('Joey Tribbiani')).toHaveTextContent('JT');
   });
 
   it('renders proper sizes', () => {
     render(<Avatar name="Joey Tribbiani" />);
-    expect(screen.getByTitle('Joey Tribbiani', { name: 'JT' })).toHaveClass(
-      'fdsAvatar--m'
-    );
+    expect(screen.getByTitle('Joey Tribbiani')).toHaveClass('fdsAvatar--m');
     cleanup();
+
     render(<Avatar name="Joey Tribbiani" size="l" />);
-    expect(screen.getByTitle('Joey Tribbiani', { name: 'JT' })).toHaveClass(
-      'fdsAvatar--l'
-    );
+    expect(screen.getByTitle('Joey Tribbiani')).toHaveClass('fdsAvatar--l');
     cleanup();
+
     render(<Avatar name="Joey Tribbiani" size="s" />);
-    expect(screen.getByTitle('Joey Tribbiani', { name: 'JT' })).toHaveClass(
-      'fdsAvatar--s'
-    );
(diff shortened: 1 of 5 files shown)


src/components/Avatar/index.tsx
```diff
@@ -13,27 +13,58 @@ const DARK_COLORS = [
...
+  PlaceholderIcon?: Icon;
+  /**
+   * Takes in a react-router `Link` reference and sets it
+   * as the base element. You may ONLY use it like the
+   * following:
+   *
+   * - `import { Link } from 'react-router'`
+   * - `Link={Link}`
+   */
+  Link?: React.ElementType;
```
sfrieson, line 61: [Could] This type could be made more narrow by using `React.ComponentType`. `React.ElementType` will accept values like `'a'` and `'div'` since they are valid elements, but they are not components. In this case `React.ComponentType` is a subset of `React.ElementType`.

mwahlmann-cbi: Fair enough

mwahlmann-cbi: ✅

sfrieson (approved):

src/components/Icon/index.tsx
```diff
@@ -0,0 +1,8 @@
+export interface IconProps {
+  color?: string;
+  size?: 'xs' | 's' | 'm' | 'l' | 'xl';
+  customSize?: number;
+}
+
+type Icon = (props: IconProps) => JSX.Element;
```
reywright, line 7: is there any convention to identify that this is a type vs a component? I thought in the other file that this was importing a component. cc @sfrieson

src/components/Avatar/index.test.tsx
```diff
@@ -37,38 +37,37 @@ describe('grabInitials', () => {
 describe('Avatar component', () => {
   it('renders name', () => {
     render(<Avatar name="Joey Tribbiani" />);
-    expect(screen.getByTitle('Joey Tribbiani', { name: 'JT' })).toBeTruthy();
+    expect(screen.getByTitle('Joey Tribbiani')).toHaveTextContent('JT');
```
reywright, line 40: I'm curious what you found out here that required you to migrate to this approach, as we're using things similar to what was used before in `src/components/Button/index.test.tsx`

reywright (commented):
Impressive @mwahlmann-cbi !

src/components/Icon/index.tsx
mwahlmann-cbi: The reason is a type and not a function (component) is because it is an abstraction. We could have created a class component and then extend it on each icon.

src/components/Avatar/index.test.tsx
reywright: Ah! The power of TypeScript, those errors were in fact legit.

src/components/Icon/index.tsx
reywright: Right. My concern is mostly convention based but I think this is fine. Contextually it makes sense now.

reywright (approved):

src/components/Icon/index.tsx
```diff
@@ -0,0 +1,8 @@
+export interface IconProps {
+  color?: string;
+  size?: 'xs' | 's' | 'm' | 'l' | 'xl';
+  customSize?: number;
```
reywright, line 4: Nice catch on making these optional 👍

PR #698, opened 2021-11-04
Migrated Badge component to TS

mwahlmann-cbi:
## Description

Migrated `Badge` component to TS. I've also added documentation for the `IconProps`.

## Checklist

- [X] All unit tests pass


src/components/Badge/index.tsx
```diff
@@ -1,10 +1,24 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import cc from 'classcat';
+import Icon from 'components/Icon';
 
-export const VALID_TYPES = ['update', 'count'];
+export const VALID_TYPES = ['update', 'count'] as const;
```
sfrieson, line 5: No requested change here, but I wonder if we'll want to start using `enum`s for things like this rather than arrays of values. There would be less extra typing for a more verbose access.

```typescript
export enum ValidTypes {
  update = 'update',
  count = 'count',
}

export interface BadgeProps {
  ...
  type?: ValidTypes;
}
```

sfrieson (approved):

src/components/Badge/index.tsx
mwahlmann-cbi: This is a good point. The only reason not to do this would be if there are any parts of the code that expect VALID_TYPES to exists and to be an array. I've seen code like that in storybook files.

mwahlmann-cbi:
> I just happened to focus on the fact that the test file changed extensions in this PR. I know in the past we haven't realized for a bit that the tests aren't picking up `.tsx` files because of the config we have set for them. I just wanted to double-check that these tests are indeed being run. Have you noticed them in the output?

I did look at the output to make sure tests are passing. In this case I didn't make any test fail to see if that was picked up or not, but did so in a previous PR.

src/components/Badge/index.tsx
reywright: There are ways to loop / map over an enum, or convert an enum to an array, but yeah it's more involved (not that we should or shouldn't do it, just building context)

reywright (approved):

PR #699, opened 2021-11-04
Enable strict mode

reywright:
## Description

Enable strict mode. Had to modify some already TypeScript converted components.

- Also added ts-migrate

## Screenshots

N/A

## Checklist

- [x] All unit tests pass


src/components/Button/index.tsx
```diff
@@ -92,8 +91,11 @@ export const Button = forwardRef(
     ref
   ) => {
     const Element = baseElement({ href, onClick: true, as: Link });
+    interface IconProps {
```
reywright, line 94: We can go back and make the decision to make these all types later if we so choose.

src/components/Button/index.stories.tsx
```diff
@@ -1,14 +1,19 @@
...
 
-const Template = (args) => <Button label="Button" {...args} />;
+const Template = ({ label = 'Button', ...args }: ButtonProps) => (
+  <Button {...args} label={label} />
+);
 
 const INVERTED_THEMES: Theme[] = ['ghost', 'outlined'];
 
-export const Primary = Template.bind({});
+export const Primary: Story<ButtonProps> = Template.bind({});
```
reywright, line 14: @mwahlmann-cbi This will be needed on your subsequent story conversions (now that strict mode is enabled)

package.json
```diff
@@ -151,6 +151,7 @@
     "stylelint-config-recommended": "^3.0.0",
     "svg-to-jsx": "^1.0.2",
     "tinycolor2": "^1.4.1",
+    "ts-migrate": "^0.1.26",
     "typescript": "4.2.2",
```
sfrieson, line 155: Just noticed this: There are some really helpful improvements in `typescript@4.3.0`, might be good to update to latest soon.

src/components/Button/index.tsx
```diff
@@ -92,8 +91,11 @@ export const Button = forwardRef(
...
+      direction: 'left' | 'right';
```
sfrieson, line 95:
```suggestion
      direction?: 'left' | 'right';

package.json
reywright: Sweet. I'll follow up with a MR to update that and a bunch of other packages.

PR #706, opened 2021-11-05
Migrated checkbox to Typescript

mwahlmann-cbi:
## Description

Migrated the checkbox component to TS

## Checklist

- [X] All unit tests pass


src/components/Checkbox/index.test.tsx
```diff
@@ -8,7 +8,7 @@ describe('Checkbox component', () => {
     const { container } = render(
       <Checkbox label="change callback" disabled name="disabled-test" />
     );
-    const input = screen.getByLabelText('change callback');
+    const input = screen.getByLabelText('change callback') as HTMLInputElement;
```
sfrieson, line 11: [Conversation] I don't like that we're losing type safety here, but I also don't have a better solution at the moment.

We _could_ extend React Test Library to offer something like `screen.getInputByLabelText`

```typescript
function getInputByLabelText (text): HTMLInputElement {
  const element = getByLabelText(text);
  if (element.nodeName !== 'INPUT') {
    throw new Error('Not an input!');
  }
  
  return element as HTMLInputElement;
}
```

We could even make that more generic to accept the expected node, but I'm wondering if that's more energy than it's worth. I think it depends on how often we need the solution and how easy it is to import and use.

What I'm trying to avoid is people using this `as X` syntax in production code and being bit by it because they didn't realize that the types actually knew something they didn't. Seeing examples like these makes people feel like they have permission to use them elsewhere even if they don't fully understand the implications.

In the example of `getLabelByText`, `input`s are not the only elements that are able to be labeled, and not all of them support `disabled`.

src/components/Checkbox/index.tsx
```diff
@@ -1,16 +1,33 @@
...
 
 import CheckEmptyIcon from 'lib/icons/react/CheckEmptyIcon';
 import CheckFilledIcon from 'lib/icons/react/CheckFilledIcon';
 import CheckIndeterminateIcon from 'lib/icons/react/CheckIndeterminateIcon';
 
-/**
- * @param {Object} props react props
- * @returns {ReactElement}
- */
+export interface CheckboxProps extends HTMLAttributes<HTMLElement | HTMLInputElement> {
```
sfrieson, line 9: I think this should only be `HTMLInputElement` since the `rest` props are spread on an `input`.

```suggestion
export interface CheckboxProps extends HTMLAttributes<HTMLInputElement> {
```

mwahlmann-cbi: Done

src/components/Checkbox/index.test.tsx
mwahlmann-cbi: Well these tests are a little bit more introspective and that's why they require the casting. Prod code shouldn't need this.

sfrieson (approved):

src/components/Checkbox/index.test.tsx
reywright: I'm going to do a little research here myself to see what people are doing here and then approve this a bit later if people are not using any alternative solutions for this.

reywright:
```suggestion
    const input = screen.getByLabelText('change callback');
```

src/components/Checkbox/index.test.tsx
```diff
@@ -24,7 +24,7 @@ describe('Checkbox component', () => {
     render(
       <Checkbox label="change callback" name="check-callback-test" onChange={changeFn} />
     );
-    const input = screen.getByLabelText('change callback');
+    const input = screen.getByLabelText('change callback') as HTMLInputElement;
```
reywright, line 27:
```suggestion
    const input = screen.getByLabelText('change callback');
```

src/components/Checkbox/index.test.tsx
```diff
@@ -43,7 +43,7 @@ describe('Checkbox component', () => {
       />
     );
 
-    const input = screen.getByLabelText('unchecking callback');
+    const input = screen.getByLabelText('unchecking callback') as HTMLInputElement;
```
reywright, line 46:
```suggestion
    const input = screen.getByLabelText('unchecking callback');
```

src/components/Checkbox/index.test.tsx
```diff
@@ -8,7 +8,7 @@ describe('Checkbox component', () => {
     const { container } = render(
       <Checkbox label="change callback" disabled name="disabled-test" />
     );
-    const input = screen.getByLabelText('change callback');
+    const input = screen.getByLabelText('change callback') as HTMLInputElement;
     expect(input.disabled).toBeTruthy();
```
reywright, line 12:
```suggestion
    expect(input).toHaveAttribute('disabled');
```

src/components/Checkbox/index.test.tsx
```diff
@@ -24,7 +24,7 @@ describe('Checkbox component', () => {
     render(
       <Checkbox label="change callback" name="check-callback-test" onChange={changeFn} />
     );
-    const input = screen.getByLabelText('change callback');
+    const input = screen.getByLabelText('change callback') as HTMLInputElement;
     expect(changeFn).not.toHaveBeenCalled();
     fireEvent.click(input);
     expect(input.checked).toBeTruthy();
```
reywright, line 30:
```suggestion
    expect(input).toBeChecked();
```

src/components/Checkbox/index.test.tsx
```diff
@@ -43,7 +43,7 @@ describe('Checkbox component', () => {
       />
     );
 
-    const input = screen.getByLabelText('unchecking callback');
+    const input = screen.getByLabelText('unchecking callback') as HTMLInputElement;
     expect(changeFn).not.toHaveBeenCalled();
     fireEvent.click(input);
     expect(input.checked).toBeFalsy();
```
reywright, line 49:
```suggestion
    expect(input).not.toBeChecked();
```

reywright (changes requested):
These suggested changes should remove the need to use `as` here, let me know if this works for you. Feel free to either accept them as is, or encode them in a commit in one shot.

src/components/Checkbox/index.test.tsx
mwahlmann-cbi: The only reason I did this. It's because otherwise the tests will have type errors. Do you prefer that? If so I can change it back.

src/components/Checkbox/index.test.tsx
```diff
@@ -8,11 +8,11 @@ describe('Checkbox component', () => {
     const { container } = render(
       <Checkbox label="change callback" disabled name="disabled-test" />
     );
-    const input = screen.getByLabelText('change callback') as HTMLInputElement;
-    expect(input.disabled).toBeTruthy();
+    const input = screen.getByLabelText('change callback');
+    expect(input).toBeDisabled();
```
reywright, line 12: nice! I was trying to search for this one.

reywright (approved):
LGTM, thanks for the code review updates.

src/components/Checkbox/index.test.tsx
reywright: Sorry didn't catch this comment. 

- Combined with the other suggested adjustments, there should be no further type errors.
- The problem is directly accessing attributes off of elements, rather than using the appropriate expectations from testing library jest-dom custom matchers: https://github.com/testing-library/jest-dom
- I've setup CI to fail if any type errors whatsoever are introduced.

PR #707, opened 2021-11-05
Migrating ZeroState to TypeScript

reywright:
## Description

Migrate component to TypeScript

## Screenshots

N//A


src/components/ZeroState/index.tsx
```diff
@@ -15,21 +21,4 @@ const ZeroState = ({ Icon, label = '', height = '0px' }) => (
...
-  Icon: PropTypes.func,
-
-  /**
-   * Text inside
-   */
-  label: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.node]),
-
-  /**
-   * Min height of section
-   */
```
sfrieson, line 31: Not sure if you're using `ts-migrate` but it unfortunately doesn't move the prop comments to the type definition.

reywright: done. Be well!

PR #731, opened 2021-11-25
Migrate Prompt to typescript.

ivanhdez28:
Changed from `PropTypes` to `interface` for props and renamed test, story and snapshot


src/components/Prompt/index.tsx
```diff
@@ -1,14 +1,34 @@
...
+  onDismiss?: () => void;
+  /** Keeps the Prompt from being rendered in a portal and renders it locally instead. */
+  disablePortal?: boolean;
+}
+
+const Prompt = (props: PromptProps): JSX.Element => (
   <Dialog
     content={props.desc ? <div id="a11y-dialog-desc">{props.desc}</div> : null}
     footerContent={
-      <Flex justify="end" wrap>
```
reywright, line 11: whoops! Glad TypeScript catches stuff like this.

PR #734, opened 2021-11-27
Migrate IconButton to typescript.

ivanhdez28:
Changed from `PropTypes` to `interface` for props and renamed test, story and snapshot.


src/components/IconButton/index.tsx
```diff
@@ -1,12 +1,46 @@
...
+export interface IconButtonProps {
+  /**
+   * Takes in a react-router `Link` reference and sets it
+   * as the base element. You may ONLY use it like the
+   * following:
+   *
+   * - `import { Link } from 'react-router'`
+   * - `Link={Link}`
+   */
+  Link?: React.ElementType;
```
reywright, line 19: Have you tested this? Also, is there maybe some sort of type definition we could pull in from react-router?

ivanhdez28: We can do something like `Link?: React.ElementType<LinkProps>;`, we'd just need `react-router`'s types, I'll try it out.

reywright: Hmmm actually now that I think about it 😅  We're really far behind in terms of react router versions. That may not be the best idea. I dunno.

PR #741, opened 2021-11-29
fix: migrate DropdownButton to typescript.

ivanhdez28:
Changed props from `PropTypes` to `interface`, wrote one unit test because there were none, can improve upon that later on.


src/components/DropdownButton/index.test.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import DropdownButton from '.';
+
+describe('DropdownButton component', () => {
+  it('renders correctly', () => {
+    const { getByText } = render(<DropdownButton>Dropdown</DropdownButton>);
```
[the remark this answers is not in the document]
ivanhdez28: there isn't really a noticeable advantage to either, it comes down to preference, i hadn't notice we had a pattern of using `screen`, will update to match that.

src/components/DropdownButton/index.test.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { screen, render } from '@testing-library/react';
+import DropdownButton from '.';
+
+describe('DropdownButton component', () => {
+  it('renders correctly', () => {
+    render(<DropdownButton>Dropdown</DropdownButton>);
+    const child = screen.getByText('Dropdown');
+    expect(child).toBeDefined();
```
reywright, line 9: [should] I'm curious if we want to use this expectation? 

https://jestjs.io/docs/expect#tobedefined
> Use .toBeDefined to check that a variable is not undefined.

What do you think about using this assertion? 
https://github.com/testing-library/jest-dom/#tobeinthedocument

I think it's more clear to someone casually glancing at this, what we're testing for if we use that assertion. We are currently using `toBeTruthy` checks but I think either just keeping that pattern, or using this above linked assertion will be better long term. What do you think?

reywright (changes requested):

src/components/DropdownButton/index.test.tsx
ivanhdez28: In my opinion, both are the same, if we already have all with `toBeTruthy` we should keep that pattern.

Normally I don't use this kind of assertions, I aim to do things like, get the text of the element and the assert against that, in this case I wanted to do `expect(child.text).toEqual("Dropdown");` but I couldn't easily get the text, and I figured I was focusing too much on something that isn't as important *right now*.

I will change `toBeDefined` to `toBeTruthy` to complete this merge.

reywright (approved):

PR #749, opened 2021-11-30
Migrate Dialog to typescript.

ivanhdez28:
Migrate Dialog to typescript.
Also refactored `noScroll.js` to add types and not use `globalThis` since it's only used in `Dialog` component and it's not a library for others to use; and removed the `CommonJS` style export in favor of `ESM`.


reywright (changes requested):
noScroll.ts ideally should not appear as a new file. We may need to pair on this (git wise) but you'd generally want to use `git mv` so that it appears as the same file, but just renamed.

reywright (approved):
LGTM! I'm not sure why github is showing noScroll as a separate file but it's preserving history for me locally.

PR #752, opened 2021-12-01
Migrate Radio to typescript.

ivanhdez28:
Migrate Radio to typescript.
- Use of `interface` instead of `PropTypes`, updated unit tests & stories.


src/components/Radio/index.stories.tsx
```diff
@@ -65,14 +65,14 @@ readingSelectedValue.parameters = {
   },
 };
 
-export const controlled = (args) => (
-  <div onChange={(e) => action('change')(e.target.value)}>
+export const controlled = (args: RadioProps): JSX.Element => (
+  <div onChange={(e) => action('change')((e.target as HTMLInputElement).value)}>
```
reywright, line 69: no change required here, but I might create an issue to remove all the uses of this `action` method.

reywright (approved):

src/components/Radio/index.stories.tsx
reywright: https://github.com/cbinsights/form-design-system/issues/753

reywright (approved):
Looks good!

PR #763, opened 2021-12-02
Migrate Popover to typescript.

ivanhdez28:
Migrate Popover to typescript.


src/components/Popover/hooks.ts
```diff
@@ -29,7 +32,11 @@ export const useDisableScroll = (disableScrollRef, isDisabled) => {
  * @param {Boolean} isActive - if the popover is currently open/active
  * @param {Boolean} closeCallback - function that closes the popover
  */
-export const useCloseOnScroll = (closeOnScrollRef, isActive, closeCallback) => {
+export const useCloseOnScroll = (
+  closeOnScrollRef: MutableRefObject<HTMLElement> | undefined,
+  isActive: boolean,
+  closeCallback: () => void
+): void => {
```
reywright, line 39: We're going to need to pair so you can teach me these weird React generics 😂

src/components/Popover/index.tsx
```diff
@@ -90,12 +155,12 @@ const Popover = React.forwardRef(
       };
     }, [isActive]);
 
-    let triggerProps = {};
-    let hoverTimeout;
+    let triggerProps: HTMLAttributes = {};
```
reywright, line 158: Nice!

src/components/Popover/util.ts
```diff
@@ -1,13 +1,19 @@
...
-export const getPopperPlacement = (position, alignment) => {
+export const getPopperPlacement = (
+  position: Position,
+  alignment: Alignment
+): Placement => {
   // Popperjs does not have a "center" placement variation. It centers by default.
   // Our component explicitly accepts a `center` value for alignment.
   const variation = !alignment || alignment === 'center' ? '' : `-${alignment}`;
-  return `${position}${variation}`;
+  return `${position}${variation}` as Placement;
```
reywright, line 18: [should] I'm generally not a fan of using `as`. Is there another way we can approach this where you don't need to cast this type?

reywright (changes requested):

reywright:
Also sorry :) I made a change to the stories file here when I upgraded Storybook, so there's a conflict in one of those files. My change should be really straightforward in any case.

src/components/Popover/hooks.ts
ivanhdez28: sure thing, it has a bit of a steep learning curve if you're not used to Generics in general but after that it's not complex at all.

src/components/Popover/util.ts
ivanhdez28: I couldn't find one, but let me try something else that I just thought of.

ivanhdez28: ended up removing this file altogether.
- removed `alignment` and `position` props, in exchange we're using `placement` since we're just passing it directly to `Popper`

reywright: This is a breaking change though, no? We're changing the public api.

ivanhdez28: that's true, you know what, let me do some more research to see what's the best approach that doesn't involve changing the public API, so we're less prone to errors.

ivanhdez28: I think the only alternative this leaves us is, doing 
``const variation: '' | `-${PopperAlignment}`...`` in `component/Popover/util.ts`, if you're ok with this, then so am i.

reywright (approved):

PR #776, opened 2021-12-07
feat(component): add isLabelUppercase to Button component

jjniczCB:
Adds isLabelUppercase boolean prop to Button component to control if the Button will apply uppercase
styling to the label.

### DIFF
--- a/src/components/Button/index.tsx
+++ b/src/components/Button/index.tsx
@@ -137,7 +137,6 @@ export const Button = forwardRef(
             breakoutLink: isBreakoutLink,
           },
           'fdsButton',
-          'fontStyle--caps',
           'display--inlineFlex',
           'rounded--all',
           'alignChild--center--center',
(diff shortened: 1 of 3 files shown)


src/components/Button/index.tsx
```diff
@@ -80,6 +83,7 @@ export const Button = forwardRef(
       disabled,
       label,
       isActive,
+      isLabelUpperCase = true,
```
ivanhdez28, line 86: i'm not entire sure the default value here should be true, maybe a different wording on the prop name.
lets see what the @reywright thinks.

looks good to me otherwise.

reywright: Our design system mirrors designs intentions, so it just depends on what they think button text should look like by default.  I think `isUpperCase` would be a nice prop here.

reywright (changes requested):
Thanks for the contribution!

I would vote no prop. The prop solves our problem quickly but probably adds way too much complexity for the amount of value it provides. If we truly want this button to be that custom, the user should just control the casing of the strings it passes.

PR #779, opened 2021-12-07
refactor(storybook): use `argTypes` instead of importing `action` from storybook

ivanhdez28:
The functionality achieved by the way we use `action` can be achieved from storybook out of the box
using `argTypes` so we're opting for that.

`src/base-styles/helperClasses/index.stories.mdx` is still pending, it can be addressed on a future pull request, or maybe even rewrite it as CSF.


reywright:
@ivanhdez28 re: our conversation offline;

I think the remaining action has little to no value on that story, and I doubt anyone would even know that example fired an action. My immediate reaction is to remove it (I've followed up on your branch with a commit doing that).

There are some different goals with the CSF format vs MDX format. One is more for an actual component. We tend to use MDX more for non component documentation. If you want to grab some time with me, we can walk through it.

PR #782, opened 2021-12-08
ci(package.json): add a single script to lint, test and typecheck

### LINKED ISSUE #751: Create a single script command (lint, test, typecheck)
The major thing we check in CI for PR's are eslint, tests, and typechecking. We currently do not check all of these things in a precommit hook locally (just so developers don't have to no-verify commit everything, and they can also just lean on CI to run these commands for them).

However, it is beneficial in instances where you KNOW something might be broken, to run these checks locally to get a faster feedback loop, and to keep your commits cleaner.

## Acceptance criteria

- Make a single command that runs lint, tests, and typechecking.
  - You'll add this to package.json scripts. There are already a bunch of combined commands in there that you can reference.
  - You can also check CI to see the exact commands that are ran, to ensure you create a combined command that encapsulates all those commands being ran.
- Name the command `check` (but I'm not too attached to that name. Feel free to brainstorm a superior name)

David-Cohen35:
fixes #751

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,8 @@
     "stats:imports": "node ./scripts/node/util/getImportStats.js",
     "type-check": "tsc",
     "semantic-release": "semantic-release",
-    "commit": "git-cz"
+    "commit": "git-cz",
+    "check": "eslint --max-warnings 0 ./src/ --ext .js,.jsx,.ts,.tsx && jest && tsc"
   },
   "peerDependencies": {
     "dayjs": "^1.9.7",


package.json
```diff
@@ -43,7 +43,8 @@
     "stats:imports": "node ./scripts/node/util/getImportStats.js",
     "type-check": "tsc",
     "semantic-release": "semantic-release",
-    "commit": "git-cz"
+    "commit": "git-cz",
+    "check": "eslint --max-warnings 0 ./src/ --ext .js,.jsx,.ts,.tsx && jest && tsc"
```
reywright, line 47: [should] I would point to the existing yarn commands. There's a lint command, a test command, and a typecheck command. The single command should encapsulate all of these, that way if we change those, this single command will automatically be right. There should be some existing examples in this document.

reywright (changes requested):

reywright:
Don't worry about changing for now; I would also say this is more of a build change than a CI change. I'll adjust that in the squash (more on that later)

reywright:
hmmm the output for this command is really odd. Nothing that you did, but it doesn't output or say anything for the lint and the typecheck commands. Weird.

PR #792, opened 2021-12-08
Refactor button background color opacity

David-Cohen35:


reywright:
I'd like to wait here a bit until we get actual opacity values from our designers.

reywright:
I'm going to close this PR for now until we have values, we can reopen it at any time.

PR #797, opened 2021-12-09
refactor(): refactor floatingAction component to TypeScript

David-Cohen35:
unsure where extra "icon" is coming from or if general TypeScript is correct


src/components/FloatingAction/index.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+// import PropTypes from 'prop-types';
+import IconButton, { THEMES } from 'components/IconButton';
+
+export const VALID_FAB_THEMES = THEMES.filter((t) => t !== 'ghost');
+
+export interface FloatingActionProps {
+  /** Used to render a FDS Icon (should obnly be used for FDS Icons) */
+  icon: any;
```
reywright, line 9: This should be uppercase. You're seeing a lowercase version (`icon`) because of this. The other version of icon you're seeing which is capitalized (`Icon`) is coming from the stories.jsx file in this same folder, will augment the props.

Screen Shot 2021-12-09 at 3 50 56 PM

reywright:
I'm going to correct this in the squash but 

1. `refactor()` isn't a valid format. The `()` has to be filled in, else you just don't use it, e.g. `refactor(scope): message` or `refactor: message`.
2. The type and scope of the commit message don't need to be repeated in the commit head. e.g. `refactor(typescript): converted FloatingAction component`.  Though I think you should use `fix(typescript): converted FloatingAction` or something like that.
3. Any TypeScript conversion of files & functions meant for the consumers of this package, should be a `fix`. (you don't have to know this yet, but: the reason for this is because semantic release will not deploy anything that isn't either a breaking change, a feature, or a fix. We need TypeScript conversions to deploy so that our consumers can pick up new type definitions).

PR #813, opened 2021-12-10
build(package.json): add node range

David-Cohen35:


reywright:
Looks good to me! Thanks!

PR #814, opened 2021-12-10
refactor(): refactor InputGroup to TypeScript

David-Cohen35:
A little stuck here - line 48 and 53 have some type errors im unsure how to solve


reywright:
@David-Cohen35 let's pair on this next week; I think I was able to solve it, but it'll be best to walk through the approach I think.

ivanhdez28:
`flexSettings` in your props is `undefined | 'shrink' | 'grow'`, and you're passing that to `getFlexSettings` function, which takes a `string[]`

src/components/InputGroup/index.tsx
```diff
@@ -31,7 +32,7 @@ export const getFlexSettings = (childCount, userSettings) => {
  * @param {Object} props react props
  * @returns {ReactElement}
  */
-const InputGroup = ({ children, flexSettings }) => {
+const InputGroup = ({ children, flexSettings }: InputGroupProps): React.ReactNode => {
```
reywright, line 35: This can probably just be `JSX.Element` as you know it always will return a single one of them?

reywright:
@David-Cohen35 first TypeScript PR completed 🙌

PR #820, opened 2021-12-13
fix: migrate Menu, MenuItem & MenuLink to typescript

### LINKED ISSUE #790: Convert Menu, MenuItem, and MenuLink components to TypeScript

ivanhdez28:
Refactor to use `interface` for `PropTypes`, update stories and unit tests.

fix #790


src/components/MenuLink/index.tsx
```diff
@@ -35,7 +21,26 @@ MenuLink.propTypes = {
...
+  children,
+  isDisabled,
+  Link,
+  ...rest
+}: MenuLinkProps): JSX.Element => (
+  <ReachMenuLink
+    className={isDisabled ? 'fdsMenuItem--disabled' : undefined}
+    onSelect={!isDisabled ? onSelect : noop}
+    aria-disabled={isDisabled}
+    as={Link}
```
ivanhdez28, line 39: having trouble with this, old jsx file had `Link` as optional and so does a story, but `ReachMenuLink` doesn't like it when i pass a Link that can be undefined.

reywright: If you pass undefined into `as`, will it work? It didn't seem like it from what I saw.

reywright: I would just define `Link` as `any` here . The definition of that prop is pretty complicated from what I saw.

ivanhdez28: yeah, plus it doesn't make sense what the library expects, it says `as: 'a' | undefined`, and in their docs it says `string | Component` where component uses the `React.forwardRef` API.

normally i'm against using `any` but this in fact looks complicated and poorly documented.

src/components/MenuLink/index.tsx
```diff
@@ -35,7 +21,28 @@ MenuLink.propTypes = {
    * - `import { Link } from 'react-router'`
    * - `Link={Link}`
    */
-  Link: PropTypes.elementType,
-};
+  Link?: any;
```
reywright, line 24: Screen Shot 2021-12-14 at 7 46 13 AM

I don't necessarily want you to do this change (at least right now, we should discuss this later) but this is one way to type this.

What I'm doing here is grabbing the prop interface of `ReachMenuLink` and then I'm accessing the typedef of `as` directly. This works.

I don't necessarily want to do this, as we have many `Link` props scattered around, and I don't want us just pulling on it in one place. Ideally we need to figure out how to type `Link` ourselves and then make sure it can be passed into theirs. But I just wanted to show you this technique. I learned about most of this yesterday, and figured out how to simplify it through executeprogram.

ivanhdez28: this is interesting.
I agree about our types, we need to figure out a way to have them all placed correctly, I will add a `TODO` so we don't lose it.

src/components/MenuItem/index.tsx
```diff
@@ -1,10 +1,25 @@
...
 
 const noop = () => {};
 
-const MenuItem = ({ onSelect, children, isDisabled, ...rest }) => (
+export interface MenuItemProps {
+  /** selection callback */
+  onSelect: () => void;
+
+  /** React children - any JSX can be used in a menu item */
+  children: Array<React.ReactNode> | React.ReactNode;
```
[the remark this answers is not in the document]
ivanhdez28: you're right, it's not needed, i got carried away by converting directly from proptypes to types

reywright (approved):
Looks good to me! Merge when ready.

Squash checklist:

1. Make sure the title is semantic
2. Keep the PR number in the title
3. The body should have "fixes #790" without quotes
4. You can leave the other bulleted commits in the body.

PR #821, opened 2021-12-13
Convert button group to type script

David-Cohen35:
I have a few questions about props being used in ButtonGroup that are not declared in GroupButton along with some other questions about how the Wrapper component works in relation to its child in GroupButton


reywright:
Sure, happy to talk!

src/components/ButtonGroup/index.tsx
```diff
@@ -1,14 +1,36 @@
...
+  buttons,
+  onChange,
+  ...restBtnGroup
+}: ButtonGroupProps): JSX.Element => (
   <div {...restBtnGroup} className="btngroup">
     {buttons.map(({ value, key, Icon, onClick, label, ...restBtn }, index) => (
       <GroupButton
         key={key || label}
-        onClick={onChange ? combine(() => onChange(label || value), onClick) : onClick}
+        onClick={onChange}
```
reywright, line 33: We'll want to restore this line back which will fix the above linting errors and should fix the tests as well?

reywright (changes requested):

src/components/ButtonGroup/index.tsx
David-Cohen35: yeah, it fixed both

reywright:
I'll fix the issue that's happening here.

reywright (approved):

src/components/ButtonGroup/GroupButton/index.tsx
```diff
@@ -1,11 +1,55 @@
...
+
+  /**
+   * Takes in a react-router `Link` reference and sets it
+   * as the base element. You may ONLY use it like the
+   * following:
+   *
+   * - `import { Link } from 'react-router'`
+   * - `Link={Link}`
+   */
+  Link?: JSX.Element;
```
ivanhdez28, line 21: in other files we have this as `Link?: React.ElementType;` or `Link?: React.ComponentType`, we should probably decide on one and stick to that.

@reywright what do you think?

reywright: @ivanhdez28 I agree, though I think my question before we standardize, is which one is correct?

ivanhdez28: `JSX.Element` is a `ReactElement` whose props and type are `any`
```
declare global {
  namespace JSX {
    interface Element extends React.ReactElement<any, any> { }
  }
}
```

```
type ElementType<P = any> =
    {
        [K in keyof JSX.IntrinsicElements]: P extends JSX.IntrinsicElements[K] ? K : never
    }[keyof JSX.IntrinsicElements] |
    ComponentType<P>;
type ComponentType<P = {}> = ComponentClass<P> | FunctionComponent<P>;
```

i feel like `ComponentType` is enough but i need to do more research on this topic to make an educated comment.

reywright: Given that this PR has been open for a week, I would bias moving forward with these changes, with the knowledge that I think as we proceed, answers around what we should standardize around, and in what situations, will be made a lot more clear (especially as we start converting files in `cbi-site` to TypeScript).

PR #823, opened 2021-12-14
fix: migrate baseElement to typescript

ivanhdez28:
rename file and add types for baseElement function.


reywright (approved):
Looks good to me! Thanks.

PR #842, opened 2021-12-15
fix(storybook): migrate storybook helpers to typescript

ivanhdez28:
Migrate files with helper functions under `util/storybook-docs` to typescript


src/util/storybook-docs/knobHelpers.ts
```diff
@@ -4,7 +4,7 @@
  * @param {Array} values list of possible prop values
  * @returns {Object} options object for storybook knobs
  */
-export const arrayToOptions = (values) =>
+export const arrayToOptions = (values: Array<any>): object =>
```
reywright, line 7: heads up: this file is going to be removed the moment we migrate away from using knobs.

reywright (approved):
Looks good!

src/util/storybook-docs/knobHelpers.ts
ivanhdez28: no harm then.

PR #853, opened 2021-12-17
style(build): updated stylelint major, configuration, and applied rules

reywright:


reywright:
@David-Cohen35 updated the stylelint configuration. I believe this should fix VSCode working with CSS properly. Do you mind pulling this down, running it locally, and giving me an approval on this if it works for you?

David-Cohen35 (approved):

PR #862, opened 2021-12-17
fix(typescript): migrate DateInput and its children components to typescript

### LINKED ISSUE #826: Convert DateInput to TypeScript

David-Cohen35:
- Migrate DateInput, NavArrows & YearAndMonthSelector to typescript
- Removed duplicated DateInputPopover, instead reused Popover.

fixes #826


src/components/DateInput/NavArrows.tsx
```diff
@@ -1,30 +1,36 @@
...
  * @returns {ReactElement}
  */
-const NavArrows = ({ onPreviousClick, onNextClick }) => (
+const NavArrows = ({ onPreviousClick, onNextClick }: NavArrowProps): JSX.Element => (
   <div className="fdsDateInput-navArrows alignChild--right--center">
     <Flex noGutters>
       <FlexItem shrink>
         <IconButton
-          onClick={() => onPreviousClick()}
+          onClick={() => onPreviousClick?.()}
```
reywright, line 26: I've never seen this pattern before 🙃 would it be cleaner to just assign a `noop` as a default parameter to these 2 parameters? That way we don't have to optional chain here?

src/components/DateInput/YearAndMonthSelector.tsx
```diff
@@ -1,30 +1,54 @@
...
-const YearAndMonthSelector = ({ date, localeUtils, onChange, startYear, endYear }) => {
-  const months = localeUtils.getMonths();
+const YearAndMonthSelector = ({
+  date,
+  localeUtils,
+  onChange,
+  startYear,
+  endYear,
+}: YearAndMonthSelectorProps): JSX.Element => {
+  const months: ReturnType<typeof localeUtils.getMonths> = localeUtils.getMonths();
```
reywright, line 34: I'm not this far with TypeScript. Can I understand a bit more of what this is doing?

src/components/DateInput/YearAndMonthSelector.tsx
```diff
@@ -1,30 +1,54 @@
...
-      form: { year, month },
-    },
-  }) => {
-    onChange(new Date(year.value, month.value));
+  const handleMonthChange: React.ChangeEventHandler<HTMLSelectElement> = (event) => {
+    onChange?.(new Date(date.getFullYear(), Number(event.target.value)));
+  };
+
+  const handleYearChange: React.ChangeEventHandler<HTMLSelectElement> = (event) => {
+    onChange?.(new Date(Number(event.target.value), date.getMonth()));
```
reywright, line 43: Same thing here. Could assigning a noop as a default param avoid the need to optional chain here?

src/components/DateInput/YearAndMonthSelector.tsx
```diff
@@ -1,30 +1,54 @@
...
   };
 
   return (
-    <form className="DayPicker-Caption border--bottom">
+    <div className="DayPicker-Caption border--bottom">
```
reywright, line 47: Could I understand why this was using `form` before, and the idea behind changing this to a `div`?

src/components/DateInput/index.tsx
```diff
@@ -29,6 +28,71 @@ export const DATE_FORMAT_MAP = {
...
+  inputRef?: MutableRefObject<HTMLInputElement>;
+
+  /** Ref to apply to popover content element */
+  popoverRef?: MutableRefObject<HTMLElement>;
+
+  /**
+   * Object accepting any valid prop from `Popover` except for `trigger` and `ref`.
+   * To apply a `ref` to the popover element, use the `popoverRef` prop.
+   */
+  popoverProps?: Omit<PopoverProps, 'ref' | 'trigger'>;
```
reywright, line 87: nice 🙂

src/components/DateInput/index.tsx
```diff
@@ -1,18 +1,17 @@
-import React, { useState } from 'react';
-import PropTypes from 'prop-types';
+import React, { MutableRefObject, useState } from 'react';
 import dayjs from 'dayjs';
 import customParseFormat from 'dayjs/plugin/customParseFormat';
 import localizedFormat from 'dayjs/plugin/localizedFormat';
 import utc from 'dayjs/plugin/utc';
 
-import DayPicker from 'react-day-picker';
+import DayPicker, { DayModifiers } from 'react-day-picker';
```
reywright, line 7: Sweet.

src/components/Popover/index.tsx
```diff
@@ -253,7 +253,11 @@ const Popover = React.forwardRef<HTMLElement, PopoverProps>(
       >
         <Popper
           innerRef={(node) => {
-            (refContent as MutableRefObject<HTMLElement>).current = node;
+            if (typeof refContent === 'function') {
+              refContent(node);
+            } else {
+              (refContent as MutableRefObject<HTMLElement>).current = node;
+            }
```
reywright, line 260: Nice, is this a type guard? I would love a walkthrough of how you came to decide that this needed to be done.

reywright (commented):
Looks good! Just some questions.

src/components/DateInput/NavArrows.tsx
ivanhdez28: that's an alternative, yes, and probably is more readable, but in the end i guess it's a matter of preference, i'm ok with either, just optional chaining seemed like an acceptable solution to me, but we can change these to `noop`, but if we're doing that, i'd suggest we declare a `noop` at `src\util` maybe, so we just pull it from there every place we need it.

src/components/DateInput/YearAndMonthSelector.tsx
ivanhdez28: `ReturnType` gives you back the return type of a function type, it's why i did `ReturnType<typeof localeUtils.getMonths> ` rather than `ReturnType<localeUtils.getMonths>, in this case i used it so i wouldn't type tuple of 12 strings manually.

if you do `type Example = ReturnType<() => string>`, you get back `string` as a type for `Example`. in this example, what's inside the angle brackets is also a type, rather than a function reference.

[docs](https://www.typescriptlang.org/docs/handbook/utility-types.html#returntypetype)

src/components/DateInput/YearAndMonthSelector.tsx
ivanhdez28: yeah, same thing applies, we could probably do something like `export const noop = (...args: any[]) => {};` and import it where we need it.

src/components/DateInput/YearAndMonthSelector.tsx
ivanhdez28: it was giving us trouble with types, it was getting complicated to type it as a form with `React.FormEvent` and then we would've had to dynamically get the values with something like `e.currentTarget.elements.namedItem('email')`, it would have to have more checks but that's where it was getting complex.

the only reason i found for using a form was to update the values getting them both at once from the form, the alternative for that was getting the value for `month` from the state when we were getting the value for `year` from the input, and viceversa.

src/components/Popover/index.tsx
ivanhdez28: the thing here is `refContent` can be either a function or a mutable ref, if it's a function it works differently, that's why there's the check for it, if it's not a function then we're sure it's a mutable ref, but typescript still doesn't like it so we have to assert it.

`forwardedRef: ((instance: HTMLElement | null) => void) | React.MutableRefObject<HTMLElement | null> | null`
forwardedRef comes from `React.forwardRef` that is wrapping the component.

src/components/DateInput/NavArrows.tsx
reywright: `src/util/noop` strategy sounds good!

src/components/DateInput/YearAndMonthSelector.tsx
reywright: oh! That's awesome!

src/components/Popover/index.tsx
reywright: hmmm. Is there any possible way to not have to use `as` here?

ivanhdez28: i'm not sure, i can't think of one, i don't know if there's something more advanced in typescript that enables this for us, i also tried `typeof` or `instanceof` and it didn't work out, but what i do know is we need it to be a `MutableRefObject` so it allows us to change the value of `current`

reywright: Ok, that's fine for now.

reywright (approved):
Looks good!

Issue #789, opened 2021-12-08
Convert all Storybook util components to TypeScript


reywright:
All the components in `src/util/storybook-docs/`

- [x] helpers (knobHelpers, useClipboard, useToggleHover)
- [ ] Badges
- [ ] #875
- [ ] #872
- [ ] ImportPath
- [ ] Info
- [ ] Layout
- [ ] Related
- [x] State
- [ ] #874
- [ ] Table

PR #864, opened 2021-12-17
fix(typescript): migrate storybook-docs Table to typescript

ivanhdez28:
migrate storybook-docs Table to typescript

re #789


src/util/storybook-docs/Table.tsx
```diff
@@ -42,13 +54,11 @@ export const TableCell = ({ children, isCSS, copy, ...props }) => {
   );
 };
 
-TableCell.propTypes = {
-  children: PropTypes.node,
-  copy: PropTypes.bool,
-  isCSS: PropTypes.bool,
-};
+export interface TableHeadLayoutProps {
+  headers?: Array<string | number>;
```
reywright, line 58: Is there a reason we would choose to use `Array<string | number>` vs `string | number[]` or (string | number)[]? Would we want to standardize our approach here?

src/util/storybook-docs/Table.tsx
```diff
@@ -58,17 +68,22 @@ const TableHeadLayout = ({ headers = [] }) => (
   </thead>
 );
 
-TableHeadLayout.propTypes = {
-  headers: PropTypes.array,
-};
+export interface TableLayoutProps {
+  rows?: Array<Array<React.ReactNode | object>>;
```
reywright, line 72: Yikes 😁 yeah there is definitely some complexity with this component

reywright (approved):
Looks good!

src/util/storybook-docs/Table.tsx
ivanhdez28: there's no difference,  `Type[]` is the shorthand for `Array<Type>`, maybe in just scenarios it would be easier to ready `Array<Array<Type>>` than `Type[][]`.

example:
image

if we're making a decision i'd say i'm leaning a bit more towards `(string | number)[]`.

src/util/storybook-docs/Table.tsx
ivanhdez28: yep, later on i'd like to have an initiative of reducing complexity on components that allow us to do so.

PR #876, opened 2021-12-21
fix(typescript): migrate storybook-docs ColorSwatch to TypeScript

David-Cohen35:
re #789


David-Cohen35:
Im not sure what to do when I'd want to make some props optional but get something like 
>   Type 'undefined' is not assignable to type 'string'.ts(2345)

Otherwise, this should work

reywright:
@David-Cohen35 if you make a field optional and you get an error about something being undefined, it either means 

1. You need to use "type guards" (I don't believe that's covered until the next executeprogram course but I or @ivanhdez28 can maybe walk you through it)
2. You need to assign a default parameter.

The thing to understand is that when a property is optional, and you assign it the type of `string`, it ultimately has the type of `string | undefined`.

So like if I do this:

```
function Blah(arg: string | undefined) {
 arg.toUpperCase()
}
```
https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAIQDYEMAWAKdAnAcwC5EBnKPGMAxAH0XABMBTYK5xgSkQG8AoRPgIA6KHACqAB0nM8AYXSlm2TnwC+QA

This will render a TypeScript error because there is no method of `toUpperCase()` on undefined. This is how TypeScript protects us.

So there's two ways to fix this. One is default params:
https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgEIBs4AtkG8BQyycUA5gPwBcyAzmFKKfgL774wCuICYwA9iDSYsAClzEyyALxEA5LOTNqGbAEo8hEqQB0YPgFUADoegBhODQgjVLIA

This can never be undefined because if it is, it will come in as an empty string.

The other way to do this is type guards:

https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAIQDYEMAWAKdAnAcwH4AuRAZyjxjAIEpEBvAKERmEWygE8AHAUzgd8BRAF4JiAOSVqtKXRaJlIgHRQ4AVV4C8AYXTl+2OqwC+zM0A

Here I check that this value absolutely is a string, before I access string methods on it. TypeScript can actually parse that I'm doing this and then it no longer flags as an error.

Happy to chat further about this or do a hangout, let me know!

reywright (approved):
I'm going to approve this, however if you're running into a situation where you wanted to type something optional, I would say to get that working first. But if not, feel free to merge away.

PR #888, opened 2021-12-21
fix(typescript): migrate storybook-docs badges to typescript

David-Cohen35:
re #789


reywright (changes requested):
I'd like to keep `null` if possible. If we need to adjust the types, that's totally fine.

David-Cohen35:
> 

two ways to do this - one is the way i just committed -- `return null as any`
and another is to set the return type to 
`JSX.Element | null`

reywright (approved):
Looks good!

PR #894, opened 2021-12-22
build(icons): fix pascalCase import statement

### LINKED ISSUE #895: Fix issue with yarn icons:export (pascal-case)
- There are a bunch of scripts that "do not run" as a part of normal CI systems.
- We recently updated a package called `pascal-case`. This introduced a change where it uses named exports, so we have to de-structure the import now, else the script breaks
- This is the easy part. The "hard" part, is that technically, I want to write a simple test that will throw when this error is not fixed. We should absolutely have some tests on these scripts, especially the ones that don't run as apart of CI, to give us coverage.

## Acceptance criteria

- Fix the bug
- Write a failing test pre-fix (I don't care much for frontend TDD)
- Loop through the different test cases.

reywright:
fixes #895


scripts/node/icons/helpers/cleanExports.js
```diff
@@ -14,32 +14,32 @@ const fs = require('fs');
...
-
-  const fileNames = filePaths.map((filePath) => {
-    let result = `${pascalCase(path.basename(filePath, ext))}.${ext}`;
-    if (ext === 'png') {
-      result = result.replace('_', '@'); // use standard PNG scale convention
-    }
-    return result;
+  filePaths.forEach((filePath) => {
+    const file = normalizePathToFile(filePath);
+    fs.renameSync(filePath, `${destination}/${file}`);
```
reywright, line 40: I did three things here things here:

- We are normalizing a single path name into a single file name. I abstracted that out into a utility function.
- I then looped over the data and then passed it to that singular utility function
- I then moved the `fileSync` into this loop, because we really don't need two.

I had to think about this for a bit but I think the resulting code is more efficient and cleaner.

scripts/node/icons/helpers/cleanExports.js
```diff
@@ -14,32 +14,31 @@ const fs = require('fs');
...
 const { sketchConfig, buildConfig } = require('../icons.config');
 
+const normalizePathToFile = (path) => {
+  const ext = path.split('.')[1];
+  let fileName = `${pascalCase(path.basename(path, ext))}.${ext}`;
+  if (ext === 'png') {
+    fileName = fileName.replace('_', '@'); // use standard PNG scale convention
+  }
+  return fileName;
+};
```
reywright, line 27: normalize function is here now. 

Note the naming:

* The function name **is** what it **does**. I don't have to make a separate comment to document what the function is doing. It's self documenting.
* Note the name of the function parameter. It is **named** what it is **doing**.
* Note the change of the variable `result` to `fileName`. It's **named** what it is **returning**.

scripts/node/icons/helpers/cleanExports.js
```diff
@@ -65,4 +64,7 @@ const cleanExports = (ext) => {
   });
 };
 
-module.exports = cleanExports;
+module.exports = {
+  cleanExports,
+  normalizePathToFile,
+};
```
reywright, line 70: I'm not sure with `module.exports` how to do both named exports and default exports, but it doesn't really matter anyway.

scripts/node/icons/helpers/cleanExports.test.ts
```diff
@@ -0,0 +1,12 @@
...
+describe('normalizing file names', () => {
+  test.concurrent.each([
+    ['/any/number/of/longer/paths/joeyTribbiani.png', 'JoeyTribbiani.png'],
+    ['/shorter/paths/monicageller.png', 'Monicageller.png'],
+    ['/any/path/CHANDLERBING.png', 'Chandlerbing.png'],
+    ['/the/path/gets/truncated/ross-gellar.png', 'RossGellar.png'],
+  ])('path of %s should return "%s"', (a, expected) => {
+    expect(normalizePathToFile(a)).toBe(expected);
+  });
+});
```
reywright, line 12: I'm trying this style of testing for the first time. I don't know how I like it yet, but a couple of thoughts as to why I'm giving it a try out:

* Rather than type out a bunch of separate expectations and type more, I can make a clean abstraction. Someone can glance at the data, the parameters, and quickly understand how all the tests are related (first argument is input, 2nd argument is the expectation).
* The `path of %s should return %s%` bit, that will render the actual values of the array in the test name itself. So we don't lose context from the test.
* You may wonder why we are not using a forEach. One reason is potentially parallelization (not that this test matters, but as a pattern in general). The other reason is because the moment any test fails with a forEach, it will either stop running any additional tests in that file, or the test suite in general. This negatively impacts actual analytics of what is broken, and also breaks DX. If I go to fix one test and then the next one breaks, that's unideal.
* I also threw on jests `concurrent` mode (because why not, if it fails we just remove the word).

scripts/node/icons/helpers/cleanExports.js
```diff
@@ -14,32 +14,31 @@ const fs = require('fs');
 const path = require('path');
 const glob = require('glob');
 const rimraf = require('rimraf');
-const pascalCase = require('pascal-case');
+const { pascalCase } = require('pascal-case');
```
reywright, line 17: This is the only thing needed to actually fix the bug. Everything else is around writing cleaner code, writing a test, and doing some experimenting while I'm doing that.

reywright (commented):
@ivanhdez28 @David-Cohen35 

Do you both mind reviewing this PR?

- I think the context of the PR and the issue itself would be interesting and educational as far as FDS goes
- I think some of the clean code techniques might be interesting to you
- I think some of the strategy around writing tests for scripts might interest you
- I definitely would recommend, when you create a PR, especially when it is complex and you know something bears explanation, to self comment your own PR. It can actually sort of pre-rebuttal things you know others may think are odd or confusing, and it serves to educate as well. Anyway take a look at the issue and PR and let me know if you have thoughts!

scripts/node/icons/helpers/cleanExports.js
reywright: One other thing: by abstracting a function out here, I was able to skip mocking `renameSync` which actually renames a file, and now I can just test the normalizePathToFile function.

scripts/node/icons/helpers/cleanExports.js
ivanhdez28: i'm not 100% sure, but i think you'd do 
```
module.exports = {
  cleanExports,
  normalizePathToFile,
  default: /* module here */
};
```

scripts/node/icons/helpers/cleanExports.test.ts
ivanhdez28: amazing! i've always wanted this in my unit tests, this is something i was familiar with when i did unit testing in .net.
my vote is on this style, looks cleaner to me, and you write less code.

ivanhdez28 (approved):
looks good to me

scripts/node/icons/helpers/cleanExports.js
reywright: unfortunately doesn't work for me.

PR #903, opened 2021-12-22
refactor: replace inline noop declarations with noop in util/index

### LINKED ISSUE #883: clean up inline noop declarations
remove inline noop declarations as default value or noop declaration in file. we rather import `noop` from `src/util/index.ts` and use that. 

comes after #882

David-Cohen35:
fixes #883


src/util/storybook-docs/ColorSwatch.tsx
```diff
@@ -22,7 +23,7 @@ export interface CopyButtonProps {
 
 const CopyButton = ({ value, copyToClipboard }: CopyButtonProps): JSX.Element => (
   <button
-    onClick={() => (copyToClipboard ? copyToClipboard((value = '')) : null)}
+    onClick={() => (copyToClipboard ? copyToClipboard((value = '')) : noop)}
```
ivanhdez28, line 26: the correct way is to set `noop` to be the default value of `copyToClipboard` in the props destructuring.

ivanhdez28 (changes requested):

ivanhdez28 (approved):
looks good

PR #905, opened 2021-12-22
build(codesandbox): add an icon to the example template

### LINKED ISSUE #904: Add an Icon to the example template

David-Cohen35:
fixes #904


reywright (approved):
Looks good!

Side question: what do you think about the change we made to add all the components into the sandbox? Do you think it's too much? Or is it helpful?

David-Cohen35:
> Looks good!
> 
> Side question: what do you think about the change we made to add all the components into the sandbox? Do you think it's too much? Or is it helpful?

They're not always necessary, but seeing a bunch of components work can provide some extra confidence that everything is working

PR #929, opened 2021-12-27
chore(deps): Revert - "update dependency lint-staged to v12.1.4"

David-Cohen35:
Reverts cbinsights/form-design-system#926


David-Cohen35:
seems to be a github issue, not from our end

PR #930, opened 2021-12-27
chore(deps): revert - update dependency react-hook-form to v7.22.5

David-Cohen35:
Reverts cbinsights/form-design-system#927


David-Cohen35:
Seems to be a github issue, not from our end

PR #934, opened 2021-12-27
fix(deps): rollback dependency postcss

David-Cohen35:
Storybook references older postcss-loader that in turn imports an older postcss-import. Rolling back postcss-import and a pinned version of postcss fixes the issue.


ivanhdez28 (approved):
sounds like a good decision to me, we rollback because master is broken, we can work on updating these deps again and make sure it all works before we merge into master.

PR #939, opened 2021-12-29
fix(styles): adjust styles to use new color function notation

### LINKED ISSUE #938: Fix stylelint Issues (from major update in PR #853)
the conversion of rgba() to rgb() has lead to invalid CSS due to a missing comma. 
ex - `rgb(var(--rgb-charcoal) 0.12);`

David-Cohen35:
- manually adjusted some css styles
- fixed format function to not use commas to match new color function notation on generated style files

fixes #938


ivanhdez28 (approved):
looks good

PR #944, opened 2021-12-30
Start converting scripts to ts-node

reywright:
Warning: I think this may "not" be working? Mostly because new changes are generated when running `yarn gen:dictionary`?


reywright:
Closing. I don't think we'll be working on this anytime soon, but it's around if we want to bring it back.

PR #946, opened 2021-12-30
chore(typescript): start fixing up excluded files

reywright:


src/components/Popover/hooks.ts
```diff
@@ -33,13 +33,13 @@ export const useDisableScroll = (
  * @param {Boolean} closeCallback - function that closes the popover
  */
 export const useCloseOnScroll = (
-  closeOnScrollRef: MutableRefObject<HTMLElement> | undefined,
+  closeOnScrollRef: RefObject<HTMLElement> | undefined,
```
ivanhdez28, line 36: changed from `MutableRefObject` to `RefObject` because `current` is **not** being reassigned.

src/components/Toaster/Toast.tsx
```diff
@@ -15,7 +15,7 @@ type ToastType = 'info' | 'warn' | 'error' | 'success' | 'progress';
 
 export interface ToastProps {
   /** JSX Content of Toast */
-  content: React.ReactNode;
+  content?: React.ReactNode;
```
ivanhdez28, line 18: rather than having it required and accepting a null, we made it optional to reduce complexity on stories and we're defaulting it to `null`.

ivanhdez28 (approved):
looks done to me, unless anyone has something to discuss.

reywright:
@ivanhdez28 Agreed. I'll leave you to merge this.

One thing here though: we'll want to open another PR to remove `tests` from being excluded, and fix all those issues as well.

Issue #940, opened 2021-12-30
Replace Enzyme with React Testing Library tests


reywright:
https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl

Enzyme is a package that used to be popular for testing React components, but has fallen out of favor, and React Testing Library has taken its place. It's a better, more maintainable way to write tests.

## Requirements

- You’ll want to have a basic understanding of the purpose of tests.
- You’ll want to have a basic understanding of how to write tests in Jest: https://www.valentinog.com/blog/jest/
- You’ll want to have a basic understanding of React Testing Library and how to write tests with it.

https://www.youtube.com/watch?v=JKOwJUM4_RM
https://www.freecodecamp.org/news/react-testing-library-tutorial-javascript-example-code/ https://www.robinwieruch.de/react-testing-library/

## Acceptance criteria

* Rewrite all tests that use Enzyme, to use React Testing Library (search for `from 'enzyme'` to quickly identify all the files this applies to)
* You may break each file out into an individual PR
* The commit names should be `test(refactor): message` or something similar to that.
* `toMatchSnapshot` is generally used in tandem with enzyme and should also be completely removed.

PR #948, opened 2022-01-03
refactor(test): remove enzyme from ButtonGroup tests in favor of RTL

ivanhdez28:
re #940


src/components/ButtonGroup/GroupButton/index.test.js
```diff
@@ -1,15 +1,14 @@
...
 const Wrapper = (props) => <div {...props} data-test="snapshot-wrapper" />;
 
 describe('GroupButton component', () => {
-  it('matches snapshot (set all props)', () => {
-    expect(
-      shallow(<GroupButton value="Feed" Icon={Icon} isActive={true} Wrapper={Wrapper} />)
-    ).toMatchSnapshot();
+  it('renders component', () => {
+    render(<GroupButton value="Feed" Icon={Icon} isActive={true} Wrapper={Wrapper} />);
+    expect(screen.getByRole('button')).toHaveTextContent('🌭');
```
reywright, line 12: It might be better to query on both the button and the name, and then the expectation itself can be simpler?

`expect(screen.getByRole('button', { name: '🌭' })).toBeInTheDocument()`

I'm not sure if this works but if it does, I think it's maybe a bit more clearer and straightforward.

reywright (changes requested):
Looks good mostly, just one potential suggestion / idea.

src/components/ButtonGroup/GroupButton/index.test.js
ivanhdez28: it does work, i didn't know this.

and you're right, it does look better, changed it to this.

reywright (approved):
Looks great! Thanks for making those changes.

PR #951, opened 2022-01-03
refactor(test): remove enzyme from Menu tests in favor of RTL

ivanhdez28:
re #940


src/components/Menu/index.test.tsx
```diff
@@ -1,48 +1,62 @@
...
+      >
+        <MenuItem onSelect={jest.fn()}>First</MenuItem>
+        <MenuItem onSelect={jest.fn()}>Second</MenuItem>
+        <MenuItem onSelect={jest.fn()}>Third</MenuItem>
       </Menu>
     );
-    expect(wrapper).toMatchSnapshot();
+    const anchor = screen.getByRole('link');
+    expect(anchor).toBeInTheDocument();
+    expect(screen.queryAllByText(/first|second|third/i).length).toStrictEqual(0);
```
reywright, line 60: Is there another way to write this last test? The way it's written, I don't really get what this is doing, and why.

reywright (changes requested):
Looks mostly good, just one area of question.

src/components/Menu/index.test.tsx
ivanhdez28: what is doing is, looking for the `MenuItem`s and asserting they're not there, because when `Menu` is disabled it just returns the `trigger`, of the top of my head I could query for each of them individually and check for them to be null.

if you have a different idea in mind, i'm all eyes.

src/components/Menu/index.test.tsx
```diff
@@ -57,6 +57,9 @@ describe('Menu component', () => {
     );
     const anchor = screen.getByRole('link');
     expect(anchor).toBeInTheDocument();
-    expect(screen.queryAllByText(/first|second|third/i).length).toStrictEqual(0);
+    // MenuItems are not expected to be rendered so we check for them to be null
+    expect(screen.queryByText('First')).toBeNull();
+    expect(screen.queryByText('Second')).toBeNull();
+    expect(screen.queryByText('Third')).toBeNull();
```
reywright, line 63: Yeah I think this is a more clear test. The other one is fancy, but this one is clear (even without the comment).

reywright (approved):

PR #952, opened 2022-01-03
refactor(test): remove enzyme from Toaster tests in favor of RTL

ivanhdez28:
re #940


src/components/Toaster/CountdownButton/index.test.tsx
```diff
@@ -1,19 +1,26 @@
...
-    expect(component).toMatchSnapshot();
+  it('renders component', () => {
+    render(<CountdownButton {...MOCK_PROPS} />);
+    expect(screen.getByTitle('Close')).toBeInTheDocument();
   });
 
+  describe('CountdownButton animation', () => {
+    it('shows a full timer stroke on mount', () => {
+      const { container } = render(<CountdownButton {...MOCK_PROPS} />);
+      const circle = container.querySelector('div > div > svg > circle');
```
reywright, line 17: My perspective is that this is probably against the principles of React Testing Library. We ideally want to test outcomes, not the makeup of the DOM itself. I would try a different approach (or happy to brainstorm with you a bit here).

reywright (changes requested):
Mostly looks good, though one part of the test may be worthy to approach in a different way.

src/components/Toaster/CountdownButton/index.test.tsx
ivanhdez28: true, 100% right what you said, we could probably pair program, there's a few things i couldn't totally grasp about the testing of animations, or maybe i'm even making it a much bigger problem than it is, but still i think could benefit from a pairing session.

reywright: Totally. Feel free to add something on my calendar for whenever you'd like.

reywright (approved):
Awesome, thanks.

PR #963, opened 2022-01-04
fix(typescript): flexSettings should be optional

### LINKED ISSUE #962: Fix optional flexSettings prop
flexSettings was initially optional but was not made so during the typeScript conversion and was now causing an error in stories

David-Cohen35:
fixes #962


David-Cohen35:
added files that werent needed - opening a new PR with correct changes

PR #964, opened 2022-01-04
fix(typescript): make flexSettings optional

### LINKED ISSUE #962: Fix optional flexSettings prop

David-Cohen35:
flexSettings was initially optional before the typescript conversion

fixes #962


ivanhdez28 (approved):
looks good

PR #966, opened 2022-01-05
fix(typescript): Icon and Link props are optional

### LINKED ISSUE #965: Fix optional props for StackedButton
Icon and Link were not required before the TS conversion.

David-Cohen35:
fixes #965

### DIFF
--- a/src/components/StackedButton/index.tsx
+++ b/src/components/StackedButton/index.tsx
@@ -13,7 +13,7 @@ export interface StackedButtonProps {
    * - `import { Link } from 'react-router'`
    * - `Link={Link}`
    */
-  Link: React.ElementType;
+  Link?: React.ElementType;
   /**
    * Controls whether the button is disabled or not. Will control the disabled
    * presentation of either an anchor or button rendered under the hood,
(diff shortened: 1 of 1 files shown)


src/components/StackedButton/index.tsx
```diff
@@ -13,15 +13,15 @@ export interface StackedButtonProps {
    * - `import { Link } from 'react-router'`
    * - `Link={Link}`
    */
-  Link: React.ElementType;
+  Link?: React.ElementType;
```
reywright, line 16: 👍

src/components/StackedButton/index.tsx
```diff
@@ -13,15 +13,15 @@ export interface StackedButtonProps {
...
   /**
    * Controls whether the button is disabled or not. Will control the disabled
    * presentation of either an anchor or button rendered under the hood,
    * but will only add a disabled attribute for buttons
    */
   disabled?: boolean;
   /**  Pass in "only" a FDS Icon reference to display it (e.g. Icon={ApproveIcon}) */
-  Icon: React.ComponentType<IconProps>;
+  Icon?: React.ComponentType<IconProps>;
```
reywright, line 24: What is the thought process behind this being optional?

David-Cohen35: it was optional before the typescript conversion - seemed like making it not optional was a mistake

reywright: If that's the case, that's the correct call I think. I sort of struggle to think when we would want to use a StackedIcon without an icon, but that's something we should think about later. Agreed.

David-Cohen35: Right, that is something to think about. I can merge this for now and we can discuss/make a new issue to make it required

PR #973, opened 2022-01-07
test(storybook): add cypress tests for storybook

David-Cohen35:
testing out a fix by adding packages with yarn instead of npm


reywright (changes requested):
Rejecting per our conversation (feel free to dismiss my review when Cypress tests, sans component testing are ready for review)

David-Cohen35:
will want to discuss folder structure - ideally to have a similar structure to jest which uses testRegex

package.json
```diff
@@ -91,6 +93,7 @@
     "@storybook/components": "6.4.9",
     "@storybook/core-events": "6.4.9",
     "@storybook/react": "6.4.9",
+    "@storybook/testing-react": "^1.2.2",
```
reywright, line 96: This can be removed

package.json
```diff
@@ -219,6 +223,7 @@
   "dependencies": {
     "@popperjs/core": "2.11.0",
     "@reach/menu-button": "0.16.2",
+    "@storybook/client-api": "6.4.10",
```
reywright, line 226: * Is this needed?
* If so, is it needed as a dependency?

cypress/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "lib": ["es5", "dom"],
+    "types": ["cypress"],
+    "jsx": "preserve"
+  },
+  "include": ["**/*.ts", "storybook/storybook.spec.ts"]
+}
```
reywright, line 9: Is this tsconfig file needed?

cypress/storybook/storybook.spec.ts
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable jest/expect-expect */
```
reywright, line 1: sure... but if we're setting up our jest regex to exclude these files, then our linter should also exclude these files ideally.

reywright (commented):
Some comments and questions.

ivanhdez28 (commented):
other than what @reywright pointed out, this looks good to me, though i'm not familiar at all with cypress so i'd merge only after he approves it.

package.json
David-Cohen35: not needed

cypress/tsconfig.json
David-Cohen35: turns out i do not

cypress/storybook/storybook.spec.ts
David-Cohen35: yeah meant to remove that - am still unsure why eslint is still saying that the test has no assertions - but it runs and passes either way

reywright (approved):
LGTM!

PR #974, opened 2022-01-10
refactor(test): remove 'getKeys' and refactor DateInput test cases to be done with static values

ivanhdez28:


reywright (approved):
Sweet! Thanks for doing this.

PR #978, opened 2022-01-11
refactor(test): remove enzyme from TextInput in favor of RTL

David-Cohen35:
re #940


ivanhdez28 (commented):
maybe get these by role too?

ivanhdez28 (approved):
looks good

PR #981, opened 2022-01-12
refactor(test): remove enzyme from SeparatorList in favor of RTL

ivanhdez28:
re #940


src/components/SeparatorList/index.test.tsx
```diff
@@ -1,21 +1,22 @@
...
+  it('renders string items', () => {
+    render(
       <SeparatorList
         separator="🐍"
         items={['why', 'did', 'it', 'have', 'to', 'be', 'snakes?']}
       />
     );
-    expect(component).toMatchSnapshot();
+    const items = screen.getAllByRole('listitem');
+    expect(items.length).toEqual(7);
```
David-Cohen35, line 15: looks good but out of curiosity is length the best way to check this? What if some are undefined?

David-Cohen35 (approved):
looks good but also curious if length is the best way to check that the items rendered

src/components/SeparatorList/index.test.tsx
reywright: Here's a snippet of code I found through googling: 

```
it("should render list of 5 fruits", () => {
  render(<FruitList />)
  const list = screen.getByRole("list", {
    name: /fruits/i,
  })
  const { getAllByRole } = within(list)
  const items = getAllByRole("listitem")
  expect(items.length).toBe(5)
})
```

reywright (changes requested):
We should look at testing the content of the list items as well potentially

src/components/SeparatorList/index.test.tsx
ivanhdez28: ohh this is interesting, I'll check it out

ivanhdez28: i changed it so we can discuss it here.

i'm not sure where you stand on this @reywright, feels like the snippet, and the blog (i went into it to look at the whole thing) does the same thing but more verbose.

this looks like a more robust lookup of elements, or what am i missing?

reywright: I guess just to state the outcome: you're adding specific text in the test. `'why', 'did', 'it'`, etc. Currently we're just making an assertion on how many list items there are. I think we would want assertions to check if the strings we're passing in are actually there.

reywright (approved):

reywright:
@ivanhdez28 definitely down to chat about some of these tests to make sure we're happy with them. But if you're ok with everything here, I think this is fine to merge.

PR #984, opened 2022-01-12
refactor(test): removes enzyme from InputGroup in favor of RTL

David-Cohen35:
re #940


ivanhdez28 (approved):
I wonder if we could be getting these two by role rather than by text, other than that it looks good to me.

ivanhdez28 (approved):
nice usage of `getByRole`.
approved.

PR #987, opened 2022-01-13
refactor(test): remove enzyme from Prompt in favor of RTL

ivanhdez28:
re #940


David-Cohen35 (approved):
looks good

reywright (approved):
Sure! I'm sort of questioning the applicability of the length check (the subsequent assertions on the text are sort of the same thing?) but that's not a change I'm requesting... just sort of thinking about it. But other than that, looks good.

PR #989, opened 2022-01-13
refactor(test): remove enzyme from FlexItem in favor of RTL

David-Cohen35:
re #940


ivanhdez28 (commented):
i'm guessing you based on the tests from `Flex`, i'm not 100% sure this is what we want but i can't think of a different approach either.

ivanhdez28 (approved):

reywright (approved):
Looks good!

PR #991, opened 2022-01-13
refactor(typescript): migrate FloatingAction tests and stories to typescript

ivanhdez28:


David-Cohen35 (approved):
looks good to me

src/components/FloatingAction/index.test.tsx
```diff
@@ -5,7 +5,7 @@ import FloatingAction from '.';
 
 describe('FloatingAction component', () => {
   it('renders component (set all props)', () => {
-    render(<FloatingAction Icon={() => null} label="delete the internet" />);
+    render(<FloatingAction Icon={() => <></>} label="delete the internet" />);
```
reywright, line 8: Is there a reason we're passing a fragment here now?

ivanhdez28: it's expecting an `Element`, the alternative is importing any Icon and pass it there, but seeing that it was passing a null, a fragment made sense.

reywright: Hmm... my sort of gut feeling is that when we create a component, specifying that it returns a `JSX.Element` makes sense. However, specifying a prop that accepts a JSX.Element seems like a bad idea unless we explicitly know we want that.

The change I'd prefer I think (happy to chat about it) is to change this back to null, and in `src/components/Icon/index.tsx`, change line `12` to say 

```
type Icon = React.ComponentType<IconProps>
```

reywright (changes requested):
Requesting some small changes on this one if that's alright.

src/components/FloatingAction/index.test.tsx
ivanhdez28: i agree with that, updated.

reywright (approved):
LGTM, thanks for addressing updates.

PR #994, opened 2022-01-14
chore(renovatebot): changes PR schedule 

### LINKED ISSUE #961: Switch up Renovate configuration
Now that we've updated most of our packages, I'd like to propose some different ideas, but I'm leaving the door open as to the exact configuration changes we want. That'll be up to whoever implements this:

- We should consider retiring the dashboard feature
- We should give Renovate more power to generate PR's for us, rather than us click checkboxes. Some ideas of cadence:
  - Have it only generate PR's every 2 weeks, on a single day (Monday, for instance)
  - Have it only generate PR's every month, on a single day (Monday, etc)

The purpose of any of the changes here, is so that we're not constantly managing package updates. Even a monthly cadence is going to be more than enough to keep our packages update, while not being a constant drain of attention.

David-Cohen35:
changes PR schedule to before 3am on the first day of the month

fixes #961


ivanhdez28 (approved):
nice

renovate.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": ["config:base", "schedule:earlyMondays"],
+  "extends": ["config:base", "schedule:before 3am on the first day of the month"],
   "timezone": "America/New_York",
   "prConcurrentLimit": 10,
```
reywright, line 4: I would remove the limit here so it just hits us all at once.

renovate.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": ["config:base", "schedule:earlyMondays"],
+  "extends": ["config:base", "schedule:before 3am on the first day of the month"],
```
reywright, line 2: I would maybe just specify first day of the month here, because in my experience Renovate may not get to spin up all the PR's it can between midnight and 3am.

reywright (changes requested):
Thanks for taking this on! Couple of requested changes.

David-Cohen35:
actually not positive this schedule config works - might have to test it out a bit

reywright:
There should be documentation (following the breadcrumbs) that shows us valid scheduling, it just requires some investigative googling.

reywright (approved):

PR #995, opened 2022-01-14
refactor(test): remove enzyme from IconInput in favor of RTL

ivanhdez28:
re #940


ivanhdez28:
may i suggest we add a role to icons? something like `role="img"`.

David-Cohen35 (approved):

reywright (approved):

reywright:
> may i suggest we add a role to icons? something like `role="img"`.

@ivanhdez28 I'm not opposed to that!

PR #1008, opened 2022-02-16
feat(icons): rename quadrant icons and add copyto icon

David-Cohen35:


reywright (changes requested):
Thanks for regenerating! It seems like something is failing here though.

ivanhdez28 (approved):
looks good

PR #1026, opened 2022-02-22
Renovate schedule changes

David-Cohen35:
this triggers every hour. If we're satisfied with the results we will change to "before 24th hour on the first day of the month"
If its still not enough, we can trigger down to the minute according to the parser


David-Cohen35:
this triggers every hour.  If we're satisfied with the results we will change to "before 24th hour on the first day of the month"
If its still not enough, we can trigger down to the minute according to the parser

reywright (approved):

PR #1027, opened 2022-02-23
Renovate schedule changes

David-Cohen35:
re #1025


reywright (approved):
Do we want to try the “every day” one first? Either way, fine with whatever changes you'd like to make here.

David-Cohen35:
i was thinking if we make all of the PR's today then there may not be any to make on Tuesday. Or are there really constant updates and we wouldn't likely have none to test Tuesday? If thats the case then ill push daily now

PR #1030, opened 2022-02-24
fix(colorswatch): fixes clicking on color did not copy to clipboard

David-Cohen35:


src/util/storybook-docs/ColorSwatch.tsx
```diff
@@ -23,7 +23,7 @@ export interface CopyButtonProps {
 
 const CopyButton = ({ value, copyToClipboard = noop }: CopyButtonProps): JSX.Element => (
   <button
-    onClick={() => copyToClipboard((value = ''))}
+    onClick={() => copyToClipboard(value ? value : '')}
```
reywright, line 26: Any ternary that goes

`x ? x : y` should be able to be simplified to `x || y`

reywright (changes requested):

src/util/storybook-docs/ColorSwatch.tsx
David-Cohen35: yeah that is beyond better - pushed the change

reywright (approved):

src/util/storybook-docs/ColorSwatch.tsx
reywright: Just go into depth, the ternary is saying 

if true ? use this value : else use this value

which is the same thing the or clause does:

use this value || use another value

In general ternaries are better when the expression is not used as one of the values. For instance:

```const width = isNavOpened ? "300px" : "100px"```

David-Cohen35: ya totally makes sense. This is a lot cleaner and easier to read

PR #1073, opened 2022-03-03
Frnd 1205 migrate dialog to radix UI

David-Cohen35:
still need to get animations working


reywright:
Couple of things that stand out immediately:

- You should be able to use the controls in Storybook to toggle the visibility of the Dialog on and off. You'll notice that the Dialog on master animates both in and out, but this new version only animates in.
- The animation should match the general speed of the current animation. This new one is a bit slower
- When migrating to an out of the box solution like this, all current code in the implementation needs to be questioned as to it's utility. I'm pretty sure you do not need the `useLayoutEffect` which locks the scroll on body for this component, as that may be provided for you with the component.
- Open the Actions tab, and try clicking the close button, and then pressing the `ESC` key. Both of these actions should theoretically produce the same output, but they don't? Please investigate. This may have to do with the way the current story is written, not entirely sure.
Screen Shot 2022-03-10 at 10 22 36 AM

David-Cohen35:
have to look into the overflow: "hidden" test more

src/components/Dialog/index.test.tsx
```diff
@@ -32,27 +32,26 @@ describe('Dialog', () => {
     expect(screen.getByText('subTitle')).toBeInTheDocument();
   });
 
-  it('tests that onDismiss gets fired onClick of close', () => {
+  it('Tests that onDismiss gets fired correctly', () => {
```
reywright, line 35: 👍

src/components/Dialog/index.test.tsx
```diff
@@ -32,27 +32,26 @@ describe('Dialog', () => {
...
     expect(screen.queryByText('content')).not.toBeInTheDocument();
-    expect(dismiss).toHaveBeenCalledTimes(1);
-  });
 
-  it('tests that pressing Esc anywhere in dialog should trigger a close', () => {
     render(dialog);
-
     fireEvent.keyDown(screen.getByTestId('dialog'), { key: 'Escape', code: 27 });
-    expect(dismiss).toHaveBeenCalledTimes(1);
+    expect(dismiss).toHaveBeenCalledTimes(2);
```
reywright, line 43: whoops 👍 this should have definitely been 2 before (one of the benefits of consuming a library, it just fixes this stuff).

src/components/Dialog/index.tsx
```diff
@@ -136,12 +136,7 @@ const Dialog = ({
                     {onDismiss && (
                       <DialogUi.Trigger asChild>
                         <div className="dialog-icon">
-                          <IconButton
-                            Icon={DenyIcon}
-                            // onClick={onDismiss}
-                            aria-label="Close"
-                            label="Close"
-                          />
+                          <IconButton Icon={DenyIcon} aria-label="Close" label="Close" />
```
reywright, line 139: Yup totally 👍

src/components/Dialog/index.test.tsx
```diff
@@ -32,27 +32,26 @@ describe('Dialog', () => {
...
   it('tests that styles are correct', () => {
     render(dialog);
 
     const computedStyles = window.getComputedStyle(screen.getByTestId('dialog'));
+
     expect(computedStyles.getPropertyValue('max-height')).toBe('1000px');
     expect(computedStyles.getPropertyValue('max-width')).toBe('1000px');
+    // expect(
+    //   window.getComputedStyle(document.documentElement).getPropertyValue('overflow')
+    // ).toBe('hidden');
```
reywright, line 55: You should know how to throw the output of the DOM here. Try to output the DOM of the previous version, and this version? It might clarify to you what this is testing.

Let me know what you find. I'm willing to have us drop this test, but just want to make sure you know what's happening here, and then try to see if something similar is happening on the new Radix Dialog implementation.

src/components/Dialog/index.tsx
```diff
@@ -101,7 +101,7 @@ const Dialog = ({
   }, [alwaysShowBorder]);
 
   const dialogNode = (
-    <DialogUi.Root defaultOpen>
+    <DialogUi.Root defaultOpen onOpenChange={onDismiss}>
```
reywright, line 104: Add

```
open={isOpen}
``` 

and get rid of `defaultOpen`. We want to make sure the dismiss logic works correctly when we pass in our own state.

David-Cohen35: This fixes the no animation when disablePortal is set to true

src/components/Dialog/index.test.tsx
David-Cohen35: Both dialog classes (the class that has the overflow property) are nested at the same level here. I'm not sure what else to look for.

src/components/Dialog/index.tsx
```diff
@@ -97,110 +92,96 @@ const Dialog = ({
...
-    </div>
+    </DialogUi.Root>
   );
 
   const transitionNode = (
     <CSSTransition timeout={200} in={isOpen} classNames="dialog" unmountOnExit>
-      <React.Fragment>
-        {disableFocusTrap ? dialogNode : <FocusTrap>{dialogNode}</FocusTrap>}
-      </React.Fragment>
+      <>{disableFocusTrap ? dialogNode : <FocusTrap>{dialogNode}</FocusTrap>}</>
```
reywright, line 175: Do we need this?

David-Cohen35: i think so - we sometimes need to disableFocusTrap

David-Cohen35: i confirmed that the prop is being used on cbi-site

reywright: right. But I guess my question, is that Radix focus traps by default. Even if that prop is being used on `cbi-site`, the focus is still going to be trapped for that instance
Screen Shot 2022-03-18 at 4 40 33 PM

David-Cohen35: ah, right good call. I'll remove it

src/components/Dialog/index.tsx
```diff
@@ -97,110 +92,96 @@ const Dialog = ({
...
-          aria-describedby="a11y-dialog-desc"
-          tabIndex={-1}
-          aria-modal="true"
-          onKeyDown={handleKeyDown}
-          style={{
-            maxWidth: width,
-            maxHeight: height,
-          }}
+        <DialogUi.Content
+          onPointerDownOutside={(event) => event.preventDefault()}
```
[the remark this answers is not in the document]
David-Cohen35: added a test to make sure that clicking the overlay doesnt close the modal - hoping that counts here

src/components/Dialog/index.tsx
David-Cohen35: the disableFocusTrap prop is now being used to toggle radix's modal prop to conditionally focus/disableFocus

reywright (approved):
LGTM! As long as it works in `cbi-site`.

PR #1075, opened 2022-03-09
Replace cypress with playwright

David-Cohen35:


David-Cohen35:
the CI test fails with: 
    browserType.launch: 
    ╔════════════════════════════════════════════════════════════╗
    ║ Host system is missing a few dependencies to run browsers.                                                     ║
    ║ Please install them with the following command:                                                                          ║
    ║                                                                                                                                                             ║
    ║     npx playwright install-deps                                                                                                          ║
    ║                                                                                                                                                             ║
    ║ <3 Playwright Team                                                                                                                           ║
    ╚════════════════════════════════════════════════════════════╝
will have to look into it further

reywright:
@David-Cohen35 it is probably as simple as executing that command they point out in the relevant .yml files before the E2E suite runs. Take a look at the deploy script and how we're running `npx semantic-release`... it's basically the exact same thing.

playwright/storybook.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+  if (testInfo.status !== testInfo.expectedStatus)
+    console.log(`Did not run as expected, ended up at ${page.url()}`);
+});
+
+test('UI Copy', async ({ page }, testInfo) => {
+  await page.goto(
+    'http://localhost:6006/?path=/story/ui-guidelines-system-messages--page'
+  );
+  const title = page.locator('#ui-copy-guidelines :text("UI Copy Guidelines")');
+  expect(title).toBeTruthy();
```
mxschmitt, line 33: a locator is always truthy and it only gets "resolved" on the page when you perform actions. In your case you want:

`await expect(title).toBeVisible()`

mxschmitt:
(There is also a `.yarn` folder checked in, which should be in the `.gitignore` file)

PR #1076, opened 2022-03-15
chore(packagejson): move radix/tooltip package to dev dependencies

David-Cohen35:
move radix/tooltip package to dev dependencies from dependencies


ivanhdez28 (approved):
looks good

PR #1078, opened 2022-03-16
feat(header): add Header as a presentational component

ivanhdez28:


David-Cohen35 (approved):

reywright:
@ivanhdez28 it looks like typechecking broke in the pipeline here.

ivanhdez28:
yea there was a small problem with how I merged the branches, fixed it una following PR.

btw, there's a small windows between when you (force) push something and when it starts the actions that allows you to merge, which is what happened to me.

PR #1082, opened 2022-03-18
test(userevent): swapping out fireEvent for userEvent when appropriate

David-Cohen35:


src/components/Checkbox/index.test.tsx
```diff
@@ -26,9 +27,9 @@ describe('Checkbox component', () => {
     );
     const input = screen.getByLabelText('change callback', { exact: false });
     expect(changeFn).not.toHaveBeenCalled();
-    fireEvent.click(input);
+    userEvent.click(input);
     expect(input).toBeChecked();
-    fireEvent.change(input, { target: input });
+    userEvent.type(input, 'change callback');
```
reywright, line 32: You see why this change is better right? It's a lot more coherent.

src/components/Dialog/index.test.tsx
```diff
@@ -32,9 +33,9 @@ describe('Dialog', () => {
 
     /* Tests that onDismiss gets fired correctly */
     /* ========================================= */
-    fireEvent.click(screen.getByLabelText('Close'));
+    userEvent.click(screen.getByLabelText('Close'));
     // Pressing esc anywhere in dialog should trigger a close
-    fireEvent.keyDown(screen.getByText('content'), { key: 'Escape', code: 27 });
+    userEvent.keyboard('{Escape}');
```
reywright, line 38: lol this is why I made this ticket. I was like "wait... this would be so much more easier with `userEvent` when you made that Dialog PR" but then I realized you were just using what was there. So that's the motivator behind why I made this ticket.

reywright (approved):
Awesome. That's all our `fireEvent` calls? If so, looks good to me!

src/components/Checkbox/index.test.tsx
David-Cohen35: yeah, the syntax here is definitely more readable

src/components/Dialog/index.test.tsx
David-Cohen35: lol ya makes sense

src/components/DateInput/index.test.tsx
```diff
@@ -18,7 +18,8 @@ describe('DateInput component', () => {
         defaultDate="2020-07-07"
       />
     );
-    fireEvent.focus(screen.getByLabelText('Date Input'));
+    // focuses the element
+    userEvent.click(screen.getByLabelText('Date Input'));
```
reywright, line 22: Are you sure this is the canonically best way to do this?

PR #1083, opened 2022-03-21
Frnd 1219 recreate fds checkbox

David-Cohen35:


src/components/Checkbox/index.tsx
```diff
@@ -1,12 +1,14 @@
...
 import cc from 'classcat';
+import * as CheckboxUi from '@radix-ui/react-checkbox';
+import * as Label from '@radix-ui/react-label';
 
 import CheckEmptyIcon from 'icons/react/CheckEmptyIcon';
 import CheckFilledIcon from 'icons/react/CheckFilledIcon';
 import CheckIndeterminateIcon from 'icons/react/CheckIndeterminateIcon';
 
-export interface CheckboxProps extends HTMLAttributes<HTMLInputElement> {
+type CheckboxExtended = Omit<HTMLAttributes<HTMLInputElement>, 'onChange'>;
```
reywright, line 10: Nice 🔥

reywright (approved):
Looks good. Are there any breaking changes here, or should this just work?

What I would suggest doing here, instead of merging this in straight away:

- Yalc this into `cbi-site`. Ask @ivanhdez28 if you have any question on how to do that.
- Test locally and see how all the checkboxes work. If that works fine, we can then merge this in.

PR #1086, opened 2022-03-25
feat(header): add Header as a presentational component

ivanhdez28:
- add `Header` as a presentational component
- `Header` internally uses `SmartInput` which is also exported
- `SmartInput` internally uses `react-select`


reywright:
While not blockers, some stuff to chip away at in this PR:

- Test Coverage
- Replace `getByTestId` where possible

src/components/Header/Logo.test.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import Logo from './Logo';
+
+describe('Logo component', () => {
+  it('should include alt property', () => {
+    const { getByAltText } = render(<Logo src="src" />);
```
reywright, line 8: Is there a reason we wouldn't access this off of `screen`?

src/components/Header/index.stories.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const ref = React.useRef<SelectInstance<OptionProps, true>>(null);
+    const [inputValue, setInputValue] = React.useState('');
+
+    const onChange = () => {};
+    const onClear = () => {};
+    const onMenuClose = () => {};
+    const onBlur = () => {};
+    const onFocus = () => {};
+    const onOptionClick = () => {};
+    const onKeyDown = () => {};
```
reywright, line 59: Is there a reason we would define all of these here, just adding the functions directly as default parameters?

src/components/SmartInput/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+} from 'react-select';
+import { customComponents, customStyles, customThemes } from './CustomSelectUtils';
+import DropdownOption from './CustomSelectComponents/DropdownOption';
+
+declare module 'react-select/dist/declarations/src/Select' {
+  export interface Props<
+    Option,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    IsMulti extends boolean,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
reywright, line 17: Is this necessary?

src/custom.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.svg' {
+  const content: string;
+  export default content;
+}
```
reywright, line 4: What is this doing?

reywright (changes requested):

src/components/Header/Logo.test.tsx
ivanhdez28: no reason, i overlooked it, updating it.

src/components/SmartInput/index.tsx
ivanhdez28: yes, i need to "augment" this type, and eslint complains about Generics not being used, but if I remove them it stops working.

src/components/Header/index.stories.tsx
ivanhdez28: no valid reason really, this is just how i built the story initially, but you're right, defaulted all of these to `noop`

src/components/SmartInput/CustomSelectComponents/CenteredHighlightedText.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  it('should include left-text if the query it is included in the suggestion', () => {
+    render(<CenteredHighlightedText {...defaultProps} query="suggestion" />);
+    expect(screen.getByTestId('left-text')).toBeInTheDocument();
+  });
+
+  it('should include centered text if the query it is included in the suggestion', () => {
+    render(<CenteredHighlightedText {...defaultProps} query="suggestion" />);
+    expect(screen.getByTestId('centered-text')).toBeInTheDocument();
```
David-Cohen35, line 27: is there another way to query for these?

src/custom.d.ts
ivanhdez28: module so that typescript recognizes `*.svg` files

src/components/SmartInput/index.tsx
reywright: I guess my question is that if you're augmenting it, then it's not unused right? Is there a way to do this without needing the ignore? Maybe additional configuration?

src/custom.d.ts
reywright: Could you elaborate further? Why does TypeScript need to recognize `.svg` files?

ivanhdez28: in `src/components/Header/index.stories.tsx` i'm import `cbi-logo.svg`, if I don't do this, it throws me this error about not finding the "module".

image

i tried to find an alternative to this, maybe something in `tsconfig.json` but i couldn't find anything, if you anybody knows of one, please let me know.

src/components/SmartInput/CustomSelectComponents/CenteredHighlightedText.test.tsx
ivanhdez28: i swear these keep popping out of nowhere, when i pushed the last commit i searched for "ByTestId" and they didn't show up.

let me fix it up.

src/components/SmartInput/index.tsx
ivanhdez28: https://react-select.com/typescript#custom-select-props its in here, it doesn't give too much info.

but if i remove it, it says that all declarations must be identical
image

doing more research on it.

ivanhdez28: i can do something like this
image

but i honestly don't like it because `formatGroupLabel` is only being put in there again to stop linter from complaining.

reywright: @cbedwinotalvaro Any thoughts here?

src/custom.d.ts
reywright: Got it. Upon searching, this seems pretty standard. Works for me!

reywright (approved):

src/components/SmartInput/CustomSelectComponents/CustomSelectedValue.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export interface SelectedValueProps {
+  item: OptionProps;
+  selectProps: Props<OptionProps, true>;
+}
+
+export const SelectedValue = ({ item, selectProps }: SelectedValueProps): JSX.Element => {
+  const { text, defaultType, isExpert, type } = item;
+  const { setCursorPosition, value, onSetSearchItems } = selectProps;
+  const itemIndex = Array.isArray(value) ? value?.findIndex((i) => i === item) : 0;
```
cbedwinotalvaro, line 14: `const itemIndex = Array.isArray(value) ? value.findIndex((i) => i === item) : 0;`

isArray would return false if the value it is null or undefined so it is not need to use the optional operator

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#using_array.isarray

src/components/SmartInput/CustomSelectComponents/CustomSelectedValue.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  const handleClick = () => setCursorPosition?.(itemIndex);
+
+  const handleDoubleClick = () => {
+    if (Array.isArray(value)) {
+      const deletedItem = value[itemIndex];
+      const newValues = value?.filter((val) => val !== deletedItem);
```
cbedwinotalvaro, line 21: `const newValues = value.filter((val) => val !== deletedItem);`

src/components/style/smartInput.css
```diff
@@ -0,0 +1,136 @@
...
+  border-radius: 3px;
+  font-size: var(--font-size-l);
+  color: var(--color-slate);
+}
+
+.logoWrapper {
+  width: 162px;
+}
+
+/* #styled-input {
```
cbedwinotalvaro, line 100: do we need this commented styles?

ivanhdez28: we don't, removed.

cbedwinotalvaro (approved):

src/components/SmartInput/index.tsx
ivanhdez28: @reywright we discuss this in a call, him and David are both on board with disabling eslint rule and keeping a comment explaining why.

PR #1120, opened 2022-04-04
fix(checkbox): migrate checkbox to radix

David-Cohen35:


src/components/Checkbox/index.test.tsx
```diff
@@ -15,57 +15,70 @@ describe('Checkbox component', () => {
...
-    expect(screen.getByLabelText('User label')).toBeInTheDocument();
+    expect(screen.queryByLabelText('Component label')).not.toBeInTheDocument();
+    expect(screen.getByText('User label')).toBeInTheDocument();
+
+    const input = screen.getByText('User label', { exact: false });
+    expect(mockOnChange).not.toHaveBeenCalled();
+    userEvent.click(input);
+    expect(input).not.toBeChecked();
+    userEvent.type(input, 'User label');
+    expect(mockOnChange).toHaveBeenCalled();
```
David-Cohen35, line 82: it couldn't find this as a label for some reason - I made a story to test it out and clicking on this user provided label does check/uncheck the checkbox but i did not see the action in the actions tab. I added the mockOnChange here to ensure that the onChange would still be called when user provides the label

reywright: @David-Cohen35 k, I would just change the one in the test to be a span.

src/components/Checkbox/index.test.tsx
```diff
@@ -15,57 +15,70 @@ describe('Checkbox component', () => {
     userEvent.click(input);
     expect(input).not.toBeChecked();
     expect(screen.getByRole('checkbox')).toHaveAttribute('disabled');
-    expect(container.querySelector('.fdsCheckable')).toHaveClass(
-      'fdsCheckable--disabled'
+    expect(container.querySelector('.fdsCheckableCheckbox')).toHaveClass(
+      'fdsCheckableCheckbox--disabled'
```
reywright, line 19: I would either checked for some disabled attribute, or just remove this test. It would be one thing if were were testing different themes, but because we're not, I would just get rid of this.

src/components/Checkbox/index.test.tsx
David-Cohen35: htmlFor is not assignable to span so it wouldnt be tied to the checkbox

David-Cohen35: if we leave it as is (with getByText instead of getByLabelText) the test is still checking that we can see the user supplied label and that the onChange callback is being fired when that label is clicked. What is the danger in leaving it like this?

reywright: ah, you're right. Yeah we can move forward with this as is, good push back.

reywright:
@David-Cohen35 this looks mostly good! Going to leave it to someone else to do a thorough review and approve.

src/components/Checkbox/index.stories.tsx
```diff
@@ -34,7 +35,7 @@ export const PreventingUserInteraction: StoryObj<CheckboxProps> = {
       <div>
         <Checkbox
           disabled
-          defaultChecked
+          checked={true}
```
ivanhdez28, line 38: not that it matters too much and i also prefer it how you have it right now, but i think the convention in the industry is that you do `<Checkbox disabled checked name="disabled-checked" label="Disabled checked" />`

ivanhdez28 (approved):
lgtm

src/components/Checkbox/index.tsx
```diff
@@ -1,67 +1,93 @@
...
   ...rest
 }: CheckboxProps): JSX.Element => {
   const IconUnchecked = CheckEmptyIcon;
   const IconChecked = indeterminate ? CheckIndeterminateIcon : CheckFilledIcon;
 
+  const [internalChecked, setChecked] = React.useState(checked ?? defaultChecked);
+
+  const handleChange = (checkedEvent: CheckboxUi.CheckedState) => {
+    if (checked === undefined) {
+      setChecked(!internalChecked);
```
cbedwinotalvaro, line 54: if the new state depends on the previous internal state it is better to use the callback provided by the setChecked function to make sure that subsequence changes are not overwritten. https://medium.com/geekculture/using-prevstate-with-react-basic-and-complex-f7f8ab4ce1b6

`setChecked((prevInternalChecked) => !prevInternalChecked)`

David-Cohen35: Nice, thanks for the article explanation!

PR #1123, opened 2022-04-06
fix(dialog): migrate dialog to radix & remove border focus style

David-Cohen35:
this includes the style changes that design-engineering requested


reywright:
Looks good to me.

src/components/Dialog/index.tsx
```diff
@@ -96,110 +90,100 @@ const Dialog = ({
 
   useEffect(() => {
     if (!alwaysShowBorder) {
-      handleResize(); // needs to fire one time immediately on mount
+      handleResize(); // needs to fire once immediately on mount
```
ivanhdez28, line 93: this will also fire on update when `alwaysShowBorder` value changes, if that's what this intended to do then i'm ok with it.

ivanhdez28 (approved):

src/components/Dialog/index.tsx
David-Cohen35: I can only assume thats whats wanted because this was here before i worked on it. I only altered the comment here

reywright: Definitely room for improvements here 😅

reywright: The thing is that if `alwaysShowBorder` is false, and someone turns it to true, we want the effect to be ran. There's no way in hooks to only run a useEffect conditionally based on the value of a boolean.

However, what you will notice here, is the conditional of `if(!alwaysShowBorder)` which actually means `handleResize` will only run if `alwaysShowBorder` is false.

So in short, the `useEffect` will rerun, but the `handleResize` function will not.

PR #1124, opened 2022-04-12
feat(nav and footer): add footer and navbar as components

clew-cbi:
This will add the header and footer that are used on the public profiles and login pages to FDS.

Also within this MR
- updated the jest config transforms so it can handle mocking SVGs.
- updates to styles flagged by stylelint


src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+} from './constants';
+import NavBarDropdown from './NavbarDropdown';
+
+const NAV_LIST = 'navList';
+const WHO_WE_SERVE = 'whoWeServe';
+const HOW_WE_HELP = 'howWeHelp';
+const WHAT_WE_OFFER = 'whatWeOffer';
+const RESOURCES = 'resources';
+
+const DEFAULT_STATE: { [key: string]: boolean } = {
```
cbedwinotalvaro, line 22: what do you think about something like this?

`type Section = 'navList' | 'whoWeServe' | 'howWeHelp' | 'whatWeOffer' | 'resources';`

`const DEFAULT_STATE: Record<Section, boolean> = {`

`const handleClick = (sectionName: Section): void => {
    setShowClass({ ...showClass, [sectionName]: !showClass[sectionName] });
  };

  const getArrowUpOrDown = (isOpen: boolean, sectionName: Section): JSX.Element => (`

in this case we have a strong type and we will throw an error if someone is trying to add a new key that it is not on the Section type

src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+export type NavBarProps = {
+  domain: string;
+  utmContext: string;
+};
+
+const NavBar = ({ domain, utmContext }: NavBarProps): JSX.Element => {
+  const [showClass, setShowClass] = useState(DEFAULT_STATE);
+
+  const handleClick = (sectionName: string): void => {
+    setShowClass({ ...showClass, [sectionName]: !showClass[sectionName] });
```
cbedwinotalvaro, line 39: If you new state depends on your previous state it is better to use the callback to set the new state to avoid overwriting changes that are executed at the same time

https://reactjs.org/docs/hooks-reference.html#functional-updates

src/components/style/index.css
```diff
@@ -18,8 +18,7 @@
 @import './cornerPosition.css';
 @import './countdownButton.css';
 @import './dateInput.css';
-@import './radio.css';
```
cbedwinotalvaro, line 21: is there a reason to remove these imports? I checked and we have some components using these styles

src/components/Navbar/index.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+    expect(screen.getByText('What We Offer')).toBeInTheDocument();
+    expect(screen.getByText('Resources')).toBeInTheDocument();
+    expect(screen.getByText('About Us')).toBeInTheDocument();
+  });
+  it('should open the dropdown on mouseover', () => {
+    render(
+      <Navbar domain="www.cbinsights.com" utmContext="utmContext=public-profiles" />
+    );
+    fireEvent.mouseOver(screen.getByText('Who We Serve'));
+    // await waitFor(() => screen.getByText('Function'))
```
cbedwinotalvaro, line 21: can we remove this?

cbedwinotalvaro (commented):
Let me know what you think about my comments

src/components/style/footer.css
```diff
@@ -0,0 +1,102 @@
+/* all fixed values are from marketing site footer */
+.FDSFootercontainer {
```
ivanhdez28, line 2: the naming convention we use for classes is like this `.fdsButton`, `.fdsButton--m`, `.fdsButton-label`.
do you think you could update them to match this?

ivanhdez28 (changes requested):

ivanhdez28:
lets us know if you need more information on any of the comments we made.

src/components/style/navbar.css
```diff
@@ -0,0 +1,501 @@
...
+.FDSnavsubMenuContainer {
+  padding: var(--space-l) 0;
+  display: none;
+  position: absolute;
+  top: 0;
+  left: 263px;
+  width: calc(100% - 263px);
+  height: 100%;
+  background-color: #6a1e6d;
+}
```
David-Cohen35, line 103: I'm not sure what's wrong here, but when I run storybook locally the submenu is not centered

src/components/style/navbar.css
```diff
@@ -0,0 +1,501 @@
...
+
+.FDSnavdropdownItem {
+  padding-left: 20px;
+  padding-right: 20px;
+  display: flex;
+  align-items: center;
+  min-height: 50px;
+  width: 263px;
```
David-Cohen35, line 111: should the width extend so that you can get the hover action without needing to hover over the text?

src/base-styles/base/documentTags.css
```diff
@@ -14,7 +14,7 @@ html {
   font-family: var(--font-family-default);
   font-size: var(--font-size-default);
   color: var(--font-color-primary);
-  text-rendering: geometricPrecision;
+  text-rendering: geometricprecision;
```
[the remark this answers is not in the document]
clew-cbi: Yes this was flagged by stylelint so I updated it.

src/components/Navbar/NavbarDropdown.tsx
```diff
@@ -0,0 +1,114 @@
...
+      <ul className={'FDSnavmenuContainer'}>
+        {content.map((menuItem) => (
+          <li
+            className={`FDSnavdropdownItem ${showDropdown ? 'FDSnavopenedSubMenu' : ''}`}
+            key={menuItem.itemName}
+          >
+            <a className={'FDSnavdropdownItemLink'} href={menuItem.url}>
+              {menuItem.itemName}
+            </a>
+            {menuItem?.subMenu?.length > 0 && (
```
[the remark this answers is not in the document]
clew-cbi: No, I'll remove it since we are controlling the data here.

src/components/style/index.css
clew-cbi: The imports I removed were duplicates.

src/components/Navbar/constants.ts
```diff
@@ -0,0 +1,169 @@
...
+  subMenuFeaturedItem?: SubMenuFeature;
+};
+
+export const WhoWeServeDropdown = (domain: string): DropdownMenu[] => [
+  {
+    itemName: 'Function',
+    url: '',
+    subMenuFeaturedItem: {
+      url: `https://${domain}.cbinsights.com/resources/case-studies/wells-fargo-case-study/`,
+      imageSrc: './Wells_Fargo_Homepage.png',
```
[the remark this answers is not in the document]
clew-cbi: Is there a certain way to handle static assets like the svgs I added in FDS?

src/components/style/navbar.css
clew-cbi: The submenu shouldn't be centered with the button.

David-Cohen35: i think i used the wrong word when i said 'centered' - i just mean it doesn't fit inside the menu

David-Cohen35: Screen Shot 2022-04-19 at 4 37 35 PM

David-Cohen35: Screen Shot 2022-04-19 at 4 38 27 PM

David-Cohen35: maybe just has to do with the svg not rendering

src/components/Navbar/constants.ts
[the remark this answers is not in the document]
David-Cohen35: the only other place i can think of doing similar things  --> /form-design-system/src/components/Header

src/components/style/navbar.css
clew-cbi: Oh good catch. I had to update the box-sizing.

src/components/Navbar/constants.ts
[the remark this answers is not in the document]
reywright: Yeah. We may (for now) need to leave the image as an argument that is passed unfortunately, just because of the way the things are setup.

cbedwinotalvaro (approved):

src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+                'fdsNavbar-navBar',
+              ])}
+            >
+              <li className={'fdsNavbar-navBarListItem'}>
+                <a className={'fdsNavbar-dropdownItemLink'}>Who We Serve</a>
+                <span className={'fdsNavbar-dropdownButton'}>
+                  {getArrowUpOrDown(showClass.whoWeServe, WHO_WE_SERVE)}
+                </span>
+                <NavBarDropdown
+                  content={WhoWeServeDropdown(site)}
```
David-Cohen35, line 136: Convention in FDS is for components to use Pascal case, and for everything else to be camel case

ivanhdez28 (approved):

David-Cohen35 (approved):

PR #1126, opened 2022-04-13
feat(tooltip): expose asChild

reywright:


src/components/Tooltip/index.tsx
```diff
@@ -14,17 +14,20 @@ export interface TooltipProps {
...
   trigger,
   message,
   position = 'bottom',
   maxWidth = 240,
+  asChild = false,
 }: TooltipProps): JSX.Element => (
   <RadixTooltip.Provider>
     <RadixTooltip.Root delayDuration={350}>
-      <RadixTooltip.Trigger asChild>{trigger}</RadixTooltip.Trigger>
+      <RadixTooltip.Trigger asChild={asChild}>{trigger}</RadixTooltip.Trigger>
```
David-Cohen35, line 30: Wouldn't this render an extra button? 
making this line `<RadixTooltip.Trigger><span>{trigger}</span></RadixTooltip.Trigger>`
would allow the tooltips to show even when passed a custom Link. Can you elaborate on the benefit of exposing asChild here?

ivanhdez28 (approved):
lgtm

src/components/Tooltip/index.tsx
ivanhdez28: we discussed what `asChild` does in a meeting.

PR #1135, opened 2022-04-25
feat(smart-input): add CustomSingleValue component

ivanhdez28:
with the possibility of a single value component, `isMulti` is not fixed to true now, and a bunch of
other types needed to be updated


cbedwinotalvaro (approved):

David-Cohen35 (approved):
lgtm

PR #1192, opened 2022-05-06
fix(smart-input): smart input keeps focus after first search

ivanhdez28:
SmartInput would lose focus on first search because React.Children.toArray changes the keys to the
inner components


cbedwinotalvaro (approved):

.tool-versions
```diff
@@ -1 +1 @@
-nodejs 16.13.0
+nodejs 16.14.2
```
reywright, line 1: This doesn't need to be handled now, but shouldn't `.tool-versions` be git ignored?

reywright: Let's focus on getting this PR out, but would definitely take a note to fix this later once header is in, if it should be ignored.

David-Cohen35 (approved):

.tool-versions
ivanhdez28: it could be, but i thought we agreed on adding it to the repo, we can remove it in a later PR if we decide on ignoring

PR #1197, opened 2022-05-17
docs(commits): add docs on how to write correct commit messages for semantic realease

ivanhdez28:


cbedwinotalvaro (approved):

contributing/commit-messages.md
```diff
@@ -0,0 +1,122 @@
+[&larr; Table of Contents](../CONTRIBUTING.md)
+
+# Commits
+
+FDS uses conventional commits to produce release versions according to semantic release.
+
+Your commit messages must have a shape like
+
+```
+<type>[optional scope]: <description>
```
David-Cohen35, line 10: should the optional scope have parenthesis around it instead of brackets?

reywright: We should also maybe start with an example, and then flesh out the structure of it. In general, we should always start simple and real world, and then break it down technically.

contributing/commit-messages.md
```diff
@@ -0,0 +1,122 @@
...
+# How to produce a new version release
+
+- fix: a commit of the type fix patches a bug in your codebase (this correlates with PATCH in Semantic Versioning).
+- feat: a commit of the type feat introduces a new feature to the codebase (this correlates with MINOR in Semantic Versioning).
+- BREAKING CHANGE: a commit that has a footer BREAKING CHANGE:, or appends a ! after the type/scope, introduces a breaking API change (correlating with MAJOR in Semantic Versioning). A BREAKING CHANGE can be part of commits of any type.
+- The rest of the types don't produce a version release when they don't contain `BREAKING CHANGE` in the description/footer or `!` in the type.
+
+# Examples
+
+## Commit message with description and breaking change footer
```
reywright, line 41: Breaking changes don't happen as often as patches and minors, so we should maybe start with examples of those first?

ivanhdez28: makes sense, let me reorder it

PR #1199, opened 2022-05-18
feat(context-menu): add ContextMenu with the same look & feel of Menu

ivanhdez28:


src/components/ContextMenu/ContextMenuContent.tsx
```diff
@@ -0,0 +1,34 @@
...
+export type ContextMenuContentProps = Pick<
+  ContextMenu.ContextMenuContentProps,
+  | 'asChild'
+  | 'children'
+  | 'onCloseAutoFocus'
+  | 'onEscapeKeyDown'
+  | 'onFocusOutside'
+  | 'onInteractOutside'
+  | 'onPointerDownOutside'
+>;
```
reywright, line 14: Nice generic usage 👍

src/components/ContextMenu/index.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+    },
+    onFocusOutside: {
+      ...hidden,
+      action: 'onFocusOutside',
+    },
+    onInteractOutside: {
+      ...hidden,
+      action: 'onInteractOutside',
+    },
+  },
```
reywright, line 46: This all seems very... redundant? I don't necessarily want us to make any function here, but I wonder if you could like show this to some online Storybook people and see if there's any improvements to this sort of boilerplate we could make?

src/components/ContextMenu/index.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+  ),
+};
+
+export default {
+  title: 'Components/ContextMenu',
+  parameters: {
+    componentSubtitle: 'Renders a context dropdown containing actions the user can take.',
+  },
+  storySort: {
+    order: ['Primary', '*'],
```
reywright, line 132: ? What is this?

reywright:
You want to make sure to export this component in that big global component file as well.

src/components/ContextMenu/index.stories.tsx
ivanhdez28: this isn't needed anymore, i must have copied it from another file that wasn't cleaned up.

cbedwinotalvaro (approved):

src/components/ContextMenu/index.stories.tsx
ivanhdez28: i can improve this later, i don't think it should be a blocker for now

src/components/ContextMenu/ContextMenuContent.tsx
ivanhdez28: 🎉

PR #1202, opened 2022-05-19
Frnd 1321 migrate Avatar to radix

David-Cohen35:


src/components/Avatar/index.tsx
```diff
@@ -99,31 +100,33 @@ const Avatar = forwardRef(
...
+        <RadixAvatar.Root
+          {...rest}
+          ref={ref}
+          role={Element === 'div' ? 'img' : undefined}
+          aria-label={ariaLabel}
+          title={cleanName || 'Placeholder Avatar'}
+          className={cc([
+            {
+              'border--focus': Element !== 'div',
+              'color--white': (DARK_COLORS as ReadonlyArray<string>).includes(bgColor),
```
ivanhdez28, line 113: can we do this differently so we don't need `as`?

ivanhdez28 (changes requested):

src/components/Avatar/index.tsx
David-Cohen35: this was here from before i worked at CBI - im thinking we leave it as is for now and then when we migrate the component to use Tailwind, well fix it up

ivanhdez28: ah, alright, sounds good to me

PR #1236, opened 2022-06-07
fix(menu-item): merge className instead of override

ivanhdez28:


ivanhdez28:
> @ivanhdez28 why would we merge? Does this mean a consumer can pass their own class to this?

yea, i was discussing this with the rest of the team, right now they're passing classes to style the bottom border and make it a fixed width, do we not want to allow any styling at all?

cbedwinotalvaro (approved):

PR #1244, opened 2022-06-15
fix(button): text color for isDestructive outlined and ghost styles

David-Cohen35:
using git bisect i found the first bad commit to be 8374684c75fb798606834e0e852c6ee5bab17a8f - "update postcss (major)". The issue did not resolve when i reverted back to the original versions of postcss. Adding css without the 'matches'  pseudo-class seems to do the trick though.

### DIFF
--- a/src/components/style/button.css
+++ b/src/components/style/button.css
@@ -346,7 +346,9 @@
   border-color: var(--color-red);
 }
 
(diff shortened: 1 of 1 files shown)


reywright:
nice!

ivanhdez28 (approved):

PR #1245, opened 2022-06-15
feat(popover): add two more align options, center-right and center-left, also alignmentOffset prop

ivanhdez28:


src/components/Popover/index.tsx
```diff
@@ -113,8 +117,20 @@ const Popover = React.forwardRef<HTMLDivElement, PopoverProps>(
...
+    const finalAlignment =
+      alignment === 'center-right'
+        ? 'start'
+        : alignment === 'center-left'
+        ? 'end'
+        : alignment;
+    const finalAlignOffset =
+      alignment === 'center-right' || alignment === 'center-left'
+        ? triggerDimensions.width / 2 + alignmentOffset
+        : alignmentOffset;
```
reywright, line 133: What do you think about making a reusable hook for this (similar to `useDisableScroll`)?

ivanhdez28: i don't think this is going to be used somewhere else, i'm also confused by 'hook', wouldn't this just be a function if we want to extract the functionality from here?

ivanhdez28: maybe something like this?

```
export const usePopoverAlignment = (
  initialAlignment: PopoverAlignment = 'start',
  initialAlignmentOffset = 0
): [ReturnType<typeof useMeasure>[0], RadixPopoverAlignment, number] => {
  const [ref, dimensions] = useMeasure();

  const alignment =
    initialAlignment === 'center-right'
      ? 'start'
      : initialAlignment === 'center-left'
      ? 'end'
      : initialAlignment;

  const alignmentOffset =
    initialAlignment === 'center-right' || initialAlignment === 'center-left'
      ? dimensions.width / 2 + initialAlignmentOffset
      : initialAlignmentOffset;

  return [ref, alignment, alignmentOffset];
};
```

// Popover/index.tsx
```
    // const [triggerRef, triggerDimensions] = useMeasure();
    // const finalAlignment =
    //   alignment === 'center-right'
    //     ? 'start'
    //     : alignment === 'center-left'
    //     ? 'end'
    //     : alignment;
    // const finalAlignmentOffset =
    //   alignment === 'center-right' || alignment === 'center-left'
    //     ? triggerDimensions.width / 2 + alignmentOffset
    //     : alignmentOffset;

    const [triggerRef, finalAlignment, finalAlignmentOffset] = usePopoverAlignment(
      alignment,
      alignmentOffset
    );
```
what do you all think?

David-Cohen35: it cleans up our code but also seems very specific to this one use case

cbedwinotalvaro: I like the idea. Even if only used in this component, creating small functions makes it easier to understand the code and, with a good name, serves as documentation.

cbedwinotalvaro (approved):

ivanhdez28:
change not needed, a different composition solves the problem this PR intended to fix.

PR #1246, opened 2022-06-16
fix(tooltip): add forward ref

David-Cohen35:


src/components/Tooltip/index.tsx
```diff
@@ -16,31 +16,36 @@ export interface TooltipProps {
...
+    { trigger, message, position = 'bottom', maxWidth = 240 }: TooltipProps,
+    forwardedRef
+  ) => {
+    const refContent = forwardedRef || useRef<HTMLDivElement>(null);
+
+    return (
+      <RadixTooltip.Provider>
+        <RadixTooltip.Root delayDuration={350}>
+          <RadixTooltip.Trigger asChild>
+            <span ref={refContent}>{trigger}</span>
```
ivanhdez28, line 30: i kind of feel that we agreed on forwarding the ref to the trigger and removing the span, didn't we?
i could be wrong.

David-Cohen35: RadixTooltip.Trigger doesn't allow a ref prop on it

ivanhdez28: gotcha, then I guess passing it to the `<span>` and using that `asChild` makes sense

ivanhdez28 (approved):

src/components/Tooltip/index.tsx
David-Cohen35: yeah idk if there's any other way to do it

PR #1280, opened 2022-07-07
FRND 1393 make callbacks trigger on popover open/close

David-Cohen35:


ivanhdez28 (approved):

src/base-styles/helperClasses/layout/width.css
```diff
@@ -5,3 +5,7 @@
 .span--50 {
   width: 50% !important;
 }
+
+.div--100 {
+  width: 100%;
+}
```
reywright, line 11: Why wouldn't we use `.span-100`?

PR #1283, opened 2022-07-15
feat(menu): allow to use "modal" property in Dropdown Menu

cbedwinotalvaro:
The "modal" property is causing performance problems in some cbi-site views that have a large volume of html elements. This change allows to send the value of the "modal" property that is required according to the implementation.

### DIFF
--- a/src/components/Menu/index.tsx
+++ b/src/components/Menu/index.tsx
@@ -4,11 +4,11 @@ import Item from './MenuItem';
 import Trigger from './MenuTrigger';
 import * as DropdownMenu from '@radix-ui/react-dropdown-menu';
 
-export type MenuProps = Omit<DropdownMenu.DropdownMenuProps, 'modal' | 'dir'>;
+export type MenuProps = Omit<DropdownMenu.DropdownMenuProps, 'dir'>;
 
-const Menu = ({ children, ...rest }: MenuProps): JSX.Element => {
+const Menu = ({ children, modal = true, ...rest }: MenuProps): JSX.Element => {
   return (
-    <DropdownMenu.Root {...rest} modal dir={'ltr'}>
+    <DropdownMenu.Root {...rest} modal={modal} dir={'ltr'}>
       {children}
     </DropdownMenu.Root>
   );


David-Cohen35 (approved):

src/components/Menu/index.tsx
```diff
@@ -4,11 +4,11 @@ import Item from './MenuItem';
 import Trigger from './MenuTrigger';
 import * as DropdownMenu from '@radix-ui/react-dropdown-menu';
 
-export type MenuProps = Omit<DropdownMenu.DropdownMenuProps, 'modal' | 'dir'>;
+export type MenuProps = Omit<DropdownMenu.DropdownMenuProps, 'dir'>;
 
-const Menu = ({ children, ...rest }: MenuProps): JSX.Element => {
+const Menu = ({ children, modal = true, ...rest }: MenuProps): JSX.Element => {
```
kchen-cbi, line 9: Can we default this to false? And do the same on `Dialog` as well? Even with `modal={false}` for both cases, the focus trap still works as far as I can tell, and for `Dialog` we still can't dismiss or interactive with anything outside the dialog, even

cbedwinotalvaro: radix uses that props to improve accessibility and I think we can set it to false only if necessary. is there other place where it is having performance issues? I would not want to leave it false by default and hide any implementation "problems" we may have in the views that use the component

kchen-cbi: `Dialog` has a prop `disableFocusTrap` that is currently used to set the `modal` value, so anywhere `Dialog` is used with `disableFocusTrap` false and a heavier number of elements on the page is going to suffer the same performance issues. The only reason I suggest it is because the accessibility difference seems minor, since as I mentioned earlier, the radix focus trapping stays active even if `modal` is false, in which case we could probably remove the `disableFocusTrap` prop entirely. it was only originally introduced because the focus trap solution we were using back then was preventing interaction with nested portalled elements, but that no longer seems to be the case (likely because we are nesting radix components).

cbedwinotalvaro: Is there another place in the application where so many html elements (> 6k) are also being created? I understand that for collections we have to send modal={false} for Menu and Dialog but I don't see a reason to leave it like that in the whole application, unless we have already identified problems in other views that we can check

kchen-cbi: Yeah I see what you mean, I'll have to look around and check. The other place I can think of is this menu in Saved Search (Advanced Search) though it's not nearly as bad:

Screen Shot 2022-07-15 at 6 19 06 PM
For Dialog though, the majority of use cases already have `disableFocusTrap` set to true (meaning `modal` is already set to false in most cases since they have an inverse relationship), so we would have to check but for sure any Dialog on collections would have issues (if `disableFocusTrap` is false -> `modal` is true).

reywright: @kchen-cbi I get your perspective... I am leaning towards `true` for a couple of reasons. Do you think we could a quick call? Want to understand the perspective from your angle and make the right call here.

ivanhdez28 (approved):

reywright (approved):

PR #1287, opened 2022-07-21
fix(popover): no longer override style prop on trigger

ivanhdez28:


ivanhdez28:
Edwin pointer out that we overlooked the fact that we're overriding `style` rather than merging.

David-Cohen35 (approved):

cbedwinotalvaro (approved):

PR #1288, opened 2022-07-26
fix(custumselectutils): adds padding left when not header style

arianamestel:
Adds padding left
Screen Shot 2022-07-26 at 4 15 23 PM


reywright (approved):

reywright:
@arianamestel thanks! Just make sure when this is squashed, that it has the `fix` prefix. If it has the `style` one, a new build will not be kicked off.

Side note: we're going to be working hard to get this back into `cbi-site` soon!

ivanhdez28 (approved):

PR #1320, opened 2022-08-02
fix(menu): replace type with interface and add subcomponents to story

David-Cohen35:


reywright (approved):

reywright:
Screen Shot 2022-08-02 at 11 32 13 AM

Looks good!

ivanhdez28 (approved):
lgtm

cbedwinotalvaro:
I don't know if I am missing something but I only see the "asChild" prop, is that right?

image
image
image

David-Cohen35:
> I don't know if I am missing something but I only see the "asChild" prop, is that right?
> 
> image image image

i dont think it is - good catch

David-Cohen35:
im not sure why `ContextMenuTrigger` shows all of the same props as `ContextMenuContent` while the formatting is the same for Menu

src/components/ContextMenu/index.tsx
```diff
@@ -3,12 +3,21 @@ import * as RadixContextMenu from '@radix-ui/react-context-menu';
...
 
-const ContextMenu = ({ children, onOpenChange }: ContextMenuProps): JSX.Element => {
+const ContextMenu = ({
+  children,
+  onOpenChange,
+  ...rest
+}: ContextMenuProps): JSX.Element => {
   return (
-    <RadixContextMenu.Root onOpenChange={onOpenChange} dir="ltr" modal>
+    <RadixContextMenu.Root {...rest} onOpenChange={onOpenChange} dir="ltr" modal>
```
cbedwinotalvaro, line 20: is there a reason to do `{...rest}` here? rest is an empty object

David-Cohen35: was just matching the format of `Menu` when we didnt know the issue was the import

cbedwinotalvaro (approved):

PR #1324, opened 2022-08-29
feat(fonts): replace font family roboto condensed with roboto

David-Cohen35:


reywright (approved):
LGTM. One thing to keep in mind: when this gets merged, we can't accommodate any other FDS changes until Roboto Condensed is purge site wide, or we revert this change. Something to keep in mind.

David-Cohen35:
> LGTM. One thing to keep in mind: when this gets merged, we can't accommodate any other FDS changes until Roboto Condensed is purge site wide, or we revert this change. Something to keep in mind.

In that case I'd want to hold off until after the freeze and we merge in the other MR

cbedwinotalvaro:
do we want to delete the `--font-family-condensed` CSS variable?

David-Cohen35:
> 

I think we do and it looks like i did; did i miss something?

reywright (changes requested):
There's some other things that probably need to be addressed in this PR.

David-Cohen35:
this PR requires a change in the `yarn gen:dictionary` script

reywright:
@David-Cohen35 

- What change does it need do you think?
- What are the linting errors?
- And... _are you sure this requires a change to the script?_ 😅

David-Cohen35:
> @David-Cohen35
> 
> * What change does it need do you think?
> * What are the linting errors introduced? Specific line numbers / errors?
> * And... _are you sure this requires a change to the script?_ 😅

1. Im not exactly sure what change it requires yet. 
2. Every line where there was an `rgb` value, it changes to an `rgba` value and causes a `Stylelint` error
Multiple, but not every line where there is a hex color there is a `Stylelint` error
It capitalizes font family properties like `roboto` and `arial` causing `Stylelint` errors
3. The first thing i think when i see this happening is did we update a package used in the script that is causing these errors - so maybe not a change in the script itself but in something the script is using

David-Cohen35:
it seems as though the errors caused by running `yarn gen:dictionary` was not caused by this PR and therefore i think this PR should be merged as is.

David-Cohen35:
i believe we did not introduce the errors caused by running `yarn gen:dictionary` here and therefore think the PR should be merged as is

reywright:
@David-Cohen35 do we want to make some of the changes we discussed the other day? Removing the font family, etc?

David-Cohen35:
> @David-Cohen35 do we want to make some of the changes we discussed the other day? Removing the font family, etc?

we certainly do!

reywright:
@David-Cohen35 sweet. Anything else?

David-Cohen35:
> @David-Cohen35 sweet. Anything else?

other than that id just raise an issue to fix the `gen: dictionary issue` in a new PR

reywright:
@David-Cohen35 

ScreenShot 2022-09-27 at 11 47 20@2x

What about this condensed CSS custom property?

David-Cohen35:
> @David-Cohen35
> 
> ScreenShot 2022-09-27 at 11 47 20@2x
> 
> What about this condensed CSS custom property?

this is in a file that gets updated when running gen:dictionary and since that causes errors i would think to leave it be for now and have it fixed when https://github.com/cbinsights/form-design-system/issues/1351 is fixed

PR #1325, opened 2022-08-31
feat(navbar): add LogoComponent prop

aiwebb:
Adds a LogoComponent prop to the NavBar so cbi-site can pass a custom logo in


cbedwinotalvaro (approved):

reywright (changes requested):
I'm going to mark this as blocked, just want to understand better what this is for, we may be able to improve the abstractions / interface here. I saw you scheduled some time with me 🙂

reywright (approved):
LGTM!

